PR #17, opened 2023-05-23
draft for status types and defintiions

paulbastian:


paulbastian:
Micha proposed to change REVOKED to INVALID, as it gives less intention whether this is a permanent state and the issuers could reflect a suspension also with a 1bit size status list as well

Sakurann (changes requested):
sorry, I am not onboard with giving the issuers the complete freedom to define their own status values. for the interop purposes, the valid and invalid values MUST always be the same. ie 0 is always valid and 1 is always invalid. (or vice versa - that I do not care). so when the bit size is bigger than 2, the first bite is always valid/invalid and from the second bit, it can be suspended, etc.

tplooker:
I'm in agreement with Kristina, the more I look at this the more I think we should have one possible enumeration of statuses rather than allowing every issuer to define their own, the question then becomes how big that set needs to be as it controls how many bits are allocated to each credential in the status list.

paulbastian:
I'm fine with defining the first bit as valid/invalid. I think providing the option for additional bits offers benefits with acceptable tradeoffs

paulbastian:
I've made some changes, please recheck @tplooker @Sakurann

Sakurann:
umm I thought we agreed to remove the concept of the "typ" attribute.. so no `revocation-list` and no `suspention-status-list`.

tplooker:
Yeah that was captured here @Sakurann https://github.com/vcstuff/draft-looker-jwt-cwt-status-list/pull/20#issuecomment-1581385983

tlodderstedt:
> umm I thought we agreed to remove the concept of the "typ" attribute.. so no `revocation-list` and no `suspention-status-list`.

But I assume the Status List Token will have a `typ` JOSE header to properly distinguish it from other JWTs, right?

Sakurann:
@tlodderstedt typo on my end. yes, JWT should be typed (`typ` claim present), the current spec also introduces a `type` body claim as far as I can see, which is what I am not in agreement with.

Sakurann:
@tplooker I am confused. so the idea is to merge this PR and then remove the "type" JWT claim in PR #20?

paulbastian:
There are separate issues already open and things are being resolved in the other PR. This one is only about introduction and renaming

draft-looker-jwt-cwt-status-list.md
```diff
@@ -91,11 +91,9 @@ The following rules apply to validating the "status" (status) claim
 
 1. The claim value MUST be a valid JSON object.
 
-2. The claim value object MUST contain a "typ" (type) attribute with a string based value that represents the type of status list referenced. The value MUST be equal to that of the "typ" attribute in the "status_list" claim for the referenced status list.
+2. The claim value object MUST contain an "idx" (index) attribute with a numberic based value that represents the index to check for status information in the status list for the current JWT. The value of this attribute MUST be a non-negative number, containing a value of zero or greater. Refer to xx for details on the validation procedure.
```
tplooker, line 94:
```suggestion
2. The claim value object MUST contain an "idx" (index) attribute with a numeric based value that represents the index to check for status information in the status list for the current JWT. The value of this attribute MUST be a non-negative number, containing a value of zero or greater. Refer to xx for details on the validation procedure.
```

Sakurann (approved):
I am a little confused with calling "VALID", "INVALID" Status List `type`, but agreed this should be merged

paulbastian:
@tplooker ready for merge

PR #49, opened 2023-08-02
chore: initial commit for ttl langugage

mprorock:
## 📑 Description
adds support for `ttl` instructions for consumers/verifiers - this is important, especially with signed lists to separate the requirement to sign and reissue from when a client or cdn needs to fetch a fresh copy of the list

### DIFF
--- a/draft-looker-oauth-jwt-cwt-status-list.md
+++ b/draft-looker-oauth-jwt-cwt-status-list.md
@@ -372,7 +372,7 @@ only.  The "Claim Name" is as defined for the CWT registry, not the
 JWT registry.  The "JWT Claim Name" is equivalent to the "Claim Name"
 in the JWT registry.
 
- - Claim Name: ttt
+ - Claim Name: ttl
  - Claim Description: Time to Live
  - JWT Claim Name: "ttl"
  - Claim Key: TBD
--- a/draft-looker-oauth-jwt-cwt-status-list.md
+++ b/draft-looker-oauth-jwt-cwt-status-list.md
@@ -92,11 +92,13 @@ The following rules apply to validating a JWT-based Status List Token. Applicati
 
 5. The JWT MAY contain an "exp" (expiration time) claim that conveys when it is considered expired by its issuer.
 
-6. The JWT MAY contain other claims.
+6. The JWT MAY contain a "ttl" (time to live) claim that conveys the maximum amount of time that the JWT may be cached by a client before a fresh copy is retrieved
 
-7. The JWT MUST be digitally signed using an asymmetric cryptographic algorithm. Relying parties MUST reject the JWT if it is using a Message Authentication Code (MAC) algorithm. Relying parties MUST reject JWTs with an invalid signature.
+7. The JWT MAY contain other claims.
 
-8. Relying parties MUST reject JWTs that are not valid in all other respects per "JSON Web Token (JWT)" {{RFC7519}}.
+8. The JWT MUST be digitally signed using an asymmetric cryptographic algorithm. Relying parties MUST reject the JWT if it is using a Message Authentication Code (MAC) algorithm. Relying parties MUST reject JWTs with an invalid signature.
+
+9. Relying parties MUST reject JWTs that are not valid in all other respects per "JSON Web Token (JWT)" {{RFC7519}}.
 
 ~~~ ascii-art
 
@@ -319,7 +321,13 @@ TBD Declare whether JWT and CWT representations can be used interchangeably by t
 TODO elaborate on risks of incorrect parsing/decoding leading to erroneous status data
 
 ## Cached and Stale status lists
-TODO consumers/Verifiers of the status list should be aware if they fetch the up-to-date data
+When consumers or verifiers of the status list fetch the data, they need
+to be aware of its up-to-date status. The 'ttl' (time-to-live) claim
+provides an essential mechanism for setting an maximum cache time for the
+fetched data. This property permits distribution of a status list to a CDN
+or other distribution mechanism while giving guidance to consumers of the
+status list on how often they need to fetch a fresh copy of the status list
+even if that status list is not expired. 
 
 ## Authorized access to the Status List
 TODO elaborate on authorization mechanisms preventing misuse and profiling as described in privacy section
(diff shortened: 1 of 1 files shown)


paulbastian:
Is that the same mechanism that we intended for `exp` or what are the intended differences?

c2bo:
We definitely need to improve the wording in the draft on this - the intention was to use the default `exp` claim to signal the freshness requirements of the issuer. Advantage of this would be that it basically comes for free if you use a standard JWT library (which should already check for expiration).

mprorock:
> We definitely need to improve the wording in the draft on this - the intention was to use the default `exp` claim to signal the freshness requirements of the issuer. Advantage of this would be that it basically comes for free if you use a standard JWT library (which should already check for expiration). Do you see a benefit in having both `exp` and `ttl` or the need for different meanings/usage?

definitely a need for both:
`exp` indicates that the issuer will reissue status no later than such and such date
`ttl` is an instruction for the client side, or the cdn side as to how often to fetch a fresh copy

These two values may be in sync in some cases, but not always.  consider the case where a jwt may need status checked frequently, but the status changes are infrequent - the isssuer should not need to re-sign a status as frequently as they wish the client to check for a new copy.  pretty common issue in the supply chain arena, and others.  
think of `exp` as the renewal date for a dns record itself, and `ttl` as the `ttl` on a dns record

c2bo:
The case that I mainly had in mind was offline-validation or restricted access to the status list for privacy reasons. In these cases the holder could carry the status list and provide status list and credential during presentation.
There would be no way for `ttl` to be adhered to / verified from the verifier side for these cases but I guess that is fine and `exp` would be the claim that is enforced and `ttl` is a policy that is basically suggested to the verifier / hosting provider of the status list?

mprorock:
> There would be no way for `ttl` to be adhered to / verified from the verifier side for these cases but I guess that is fine and `exp` would be the claim that is enforced and `ttl` is a policy that is basically suggested to the verifier / hosting provider of the status list?

exactly - that is why `exp` is the more important one, but adding `ttl` as optional covers the more connected and cdn type use cases

draft-looker-oauth-jwt-cwt-status-list.md
```diff
@@ -350,7 +358,26 @@ TODO evaluate deifnition of Status List Provider?
...
+addition to both registries.
+
+The "Claim Description", "Change Controller" and "Specification
+Documents" are common and equivalent for the JWT and CWT registries.
+The "Claim Key" and "Claim Value Types(s)" are for the CWT registry
+only.  The "Claim Name" is as defined for the CWT registry, not the
+JWT registry.  The "JWT Claim Name" is equivalent to the "Claim Name"
+in the JWT registry.
+
+ - Claim Name: ttt
```
mprorock, line 375:
```suggestion
 - Claim Name: ttl
```

OR13 (approved):

paulbastian:
As we got the the caching information integrated with the HTTP Cache Control over HTTP GET in the specification, the question remains whether the duplicate information inside the Status List serves a purpose

OR13:
If you duplicate the information, the list can be delivered by the holder in an internet denied environment.

Assuming the verifier has network, or had network, might make sense for some use cases... Other use cases might assume burnt in trust roots and no network ever.

c2bo:
> If you duplicate the information, the list can be delivered by the holder in an internet denied environment.
> 
> Assuming the verifier has network, or had network, might make sense for some use cases... Other use cases might assume burnt in trust roots and no network ever.

Perhaps I am missing something but in my mind the caching information is not relevant for the verifier, only for the system fetching the updates (if they are separate entities). The verifier has `iat` and `exp` to figure out if status list should be trusted/used. The entity fetching the updates needs the information when to try fetching again for possible updates.

Would you be up for a short call on this topic so we can better understand your point of view?

paulbastian:
As we integrated a mechanism to enable Caching over the HTTP Cache-Control mechanism, we dot not see a need to add or duplicate this information to the data of the Status List Token.
Pending close without any further feedback.

paulbastian (commented):
From our latest thoughts, we are again considering to integrate this: https://github.com/vcstuff/draft-ietf-oauth-status-list/issues/96

draft-looker-oauth-jwt-cwt-status-list.md
```diff
@@ -92,11 +92,13 @@ The following rules apply to validating a JWT-based Status List Token. Applicati
 
 5. The JWT MAY contain an "exp" (expiration time) claim that conveys when it is considered expired by its issuer.
 
-6. The JWT MAY contain other claims.
+6. The JWT MAY contain a "ttl" (time to live) claim that conveys the maximum amount of time that the JWT may be cached by a client before a fresh copy is retrieved
```
tplooker, line 95:
```suggestion
6. The JWT MAY contain a "ttl" (time to live) claim that conveys the maximum amount of time, in seconds, that the JWT may be cached by a client before a fresh copy is retrieved. The value of the claim MUST be a positive number.
```

OR13: Why are we creating a new claim, that does basically the same thing as `exp` ?

draft-looker-oauth-jwt-cwt-status-list.md
```diff
@@ -92,11 +92,13 @@ The following rules apply to validating a JWT-based Status List Token. Applicati
...
 
-7. The JWT MUST be digitally signed using an asymmetric cryptographic algorithm. Relying parties MUST reject the JWT if it is using a Message Authentication Code (MAC) algorithm. Relying parties MUST reject JWTs with an invalid signature.
+7. The JWT MAY contain other claims.
```
OR13, line 97:
```suggestion
```

OR13: ^ remove throwaway sentence

draft-looker-oauth-jwt-cwt-status-list.md
```diff
@@ -350,7 +358,26 @@ TODO evaluate deifnition of Status List Provider?
...
+addition to both registries.
+
+The "Claim Description", "Change Controller" and "Specification
+Documents" are common and equivalent for the JWT and CWT registries.
+The "Claim Key" and "Claim Value Types(s)" are for the CWT registry
+only.  The "Claim Name" is as defined for the CWT registry, not the
+JWT registry.  The "JWT Claim Name" is equivalent to the "Claim Name"
+in the JWT registry.
+
+ - Claim Name: ttl
```
OR13, line 375: prefer to not register this claim... just make `exp` required.

draft-looker-oauth-jwt-cwt-status-list.md
paulbastian: you actually approved the same idea half a year ago? The idea is that Issuers/Status List Providers may not want to resign the whole Status List if nothing has changed, therefore `ttl` communicates the update interval of the Issuer, so that RP knows how often to check for updates and `exp` communicates "do not use this list after this point of time"

paulbastian:
pending close in favor for #101 , too many merge conflicts here

Issue #39, opened 2023-07-11
Use zlib instead of Gzip?


paulbastian:
Gzip is for files and uses zlib internally

paulbastian:
After some research to the difference between gzip, zlib and Deflate, I would propose to consider zlib instead Gzip:
- both rely on Deflate compression algorithm
- Gzip is more focused on files
- Gzip contains more metadata that we do not need as they are redundant in JWT or deprecated (modifytime, content type)
- header is a little smaller

bc-pi:
(apologies in advance for any misunderstanding/confusion around the relationship between gzip, zlib, and Deflate) but FWIW JWE uses DEFLATE/RFC1951 directly for its optional compression https://datatracker.ietf.org/doc/html/rfc7516#section-4.1.3 (and there's one example test in https://www.rfc-editor.org/rfc/rfc7520.html#section-5.9.4), which might be an option here.

paulbastian:
add recommendation for BEST_COMPRESSION mode

paulbastian:
Solved in #80

PR #80, opened 2023-11-01
switch to deflate + zlib instead of gzip

### LINKED ISSUE #67: Slight difference in compression result
I'm working on a TypeScript implementation of this spec, and have the encoding / decoding of the status list implemented. It seems to work based on the example in the spec and the example python scripts, with one slight difference.

When I encode either the 1 or 2 bit example it will output one character difference `_`, while mine outputs an `A`. The decoding seems to work fine, and maybe this is a gzip thing, but I'm wondering if the examples use certain gzip features to get this output?

I've adapted my implementation to use compression level 9 (default for python `gzip` lib) and set the mtime to the same value

The input byte array is the same as for the python implementation and the examples in the spec, so it's really something that is happening during the gzip encoding (the output byte array here is different)

```
H4sIAMo_jGQCA9u5GABc9QE7AgAAAA
H4sIAMo_jGQCAzvp8hMAZLRLMQMAAAA
```

```
H4sIAMoAjGQCA9u5GABc9QE7AgAAAA
H4sIAMoAjGQCAzvp8hMAZLRLMQMAAAA
```

any idea what is going on here? 

The implementation and tests can be found here: https://github.com/TimoGlastra/jwt-status-list-ts

c2bo:
Closes #67 and #39.

## 📑 Description
Switches to zlib instead of gzip encoding and adds the recommendation to use the best available compression.

## Preview Link
[click here for rendered preview of PR](https://github.com/<repo-name>/<branch-name>/<draft-name>.html)

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -462,6 +462,11 @@ for their valuable contributions, discussions and feedback to this specification
 # Document History
 {:numbered="false"}
 
+-01
+
+* Changing compression from gzip to zlib
+* Change typo in Status List Token sub claim description
+
 -00
 
 * Initial draft after working group adoption
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -131,7 +131,7 @@ Each status of a Referenced Token MUST be represented with a bit size of 1,2,4, 
 
 1. The overall Status List is encoded as a byte array. Depending on the bitsize, each byte corresponds to 8/(#bit-size) statuses (8,4,2, or 1). The status of each Referenced Token is identified using the index that maps to one or more specific bits within the byte array. The index starts counting at 0 and ends with "size" - 1 (being the last valid entry). The bits within an array are counted from least significant bit "0" to the most significant bit ("7"). All bits of the byte array at a particular index are set to a status value.
 
-2. The complete byte array is compressed using the "DEFLATE" {{RFC1951}} compression method and stored using the "ZLIB" {{RFC1950}} data format. Implementations SHOULD use the highest compression level available.
+2. The complete byte array is compressed using the "DEFLATE" {{RFC1951}} compression method and stored using the "ZLIB" {{RFC1950}} data format. Implementations are RECOMMENDED to use the highest compression level available.
 
 3. The result of the compression is then base64url-encoded, as defined in Section 2 of {{RFC7515}}.
 
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -27,7 +27,8 @@ normative:
   RFC7519: RFC7519
   RFC8392: RFC8392
   RFC3986: RFC3986
-  RFC1952: RFC1952
+  RFC1950: RFC1950
+  RFC1951: RFC1951
   RFC7515: RFC7515
   RFC6125: RFC6125
   RFC9110: RFC9110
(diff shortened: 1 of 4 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -130,9 +131,9 @@ Each status of a Referenced Token MUST be represented with a bit size of 1,2,4,
 
 1. The overall Status List is encoded as a byte array. Depending on the bitsize, each byte corresponds to 8/(#bit-size) statuses (8,4,2, or 1). The status of each Referenced Token is identified using the index that maps to one or more specific bits within the byte array. The index starts counting at 0 and ends with "size" - 1 (being the last valid entry). The bits within an array are counted from least significant bit "0" to the most significant bit ("7"). All bits of the byte array at a particular index are set to a status value.
 
-2. The complete byte array is compressed using gZIP {{RFC1952}}.
+2. The complete byte array is compressed using the "DEFLATE" {{RFC1951}} compression method and stored using the "ZLIB" {{RFC1950}} data format. Implementations SHOULD use the highest compression level available.
```
paulbastian, line 134:
```suggestion
2. The complete byte array is compressed using the "DEFLATE" {{RFC1951}} compression method and stored using the "ZLIB" {{RFC1950}} data format. Implementations are RECOMMENDED to use the highest compression level available.
```

paulbastian (approved):

tplooker (approved):

draft-ietf-oauth-status-list.md
```diff
@@ -130,9 +131,9 @@ Each status of a Referenced Token MUST be represented with a bit size of 1,2,4,
 
 1. The overall Status List is encoded as a byte array. Depending on the bitsize, each byte corresponds to 8/(#bit-size) statuses (8,4,2, or 1). The status of each Referenced Token is identified using the index that maps to one or more specific bits within the byte array. The index starts counting at 0 and ends with "size" - 1 (being the last valid entry). The bits within an array are counted from least significant bit "0" to the most significant bit ("7"). All bits of the byte array at a particular index are set to a status value.
 
-2. The complete byte array is compressed using gZIP {{RFC1952}}.
+2. The complete byte array is compressed using the "DEFLATE" {{RFC1951}} compression method and stored using the "ZLIB" {{RFC1950}} data format. Implementations are RECOMMENDED to use the highest compression level available.
```
bc-pi, line 134: I'm not sure I understand "compressed using the "DEFLATE" {{RFC1951}} compression method and stored using the "ZLIB" {{RFC1950}} data format" and not sure I understand well enough to even ask appropriate questions. Does this say to compress twice, once with DEFLATE only and once with ZLIB? That sounds silly but could easily be read that way. Or like what does "stored" mean in this context? Stored where? This should be about data format for exchange between parties. Whereas storage typically implies something internal (which wouldn't be in scope for the interchange format). I would expect this draft to clearly specify one compression method. This reads like two compressions. The code below suggests that ZLIB is being used - shouldn't the text reflect that? I know zlib does use DEFLATE internally and that is confusing things but indicating them both with an "and" is further confusing.

paulbastian: So Zlib is using Deflate as the compression method internally and zlib is more of a data format wrapping the result. "stored" may indeed be confusing here, is "encoded" better? Do you think its better not to reference Deflate as its referenced by Zlib RFC?

Some implementations may not offer the choice for the compression level, so I think that was the idea to not mandate the level of compression. Although we don't have extensive experience with many programming languages yet.

bc-pi: The wording is tricky (for me anyway!) but maybe something like this kinda following the language example from HTTP conveys the intent here more clearly: 

```suggestion
2. The complete byte array is compressed using DEFLATE {{RFC1951}} inside the ZLIB {{RFC1950}} data format. Implementations are RECOMMENDED to use the highest compression level available.
```

maybe...

bc-pi: This PR has been merged already so I created issue #94

PR #88, opened 2023-12-12
add access token example

### LINKED ISSUE #82: IETF 118: mention usecase for accesstokens
Aaron says that there are zero mentions of the term access token and proposes to mention JWT accesstokens as a possible usecase

c2bo:
Closes #82 

## 📑 Description
Adds access token as an example use-case for status list.

## Preview Link
[click here for rendered preview of PR](https://vcstuff.github.io/draft-ietf-oauth-status-list/c2bo/access-token/draft-ietf-oauth-status-list.html)


draft-ietf-oauth-status-list.md
```diff
@@ -44,7 +46,7 @@ The status list data structures themselves are also represented as JWTs or CWTs.
...
+JSON Web Tokens (JWTs) {{RFC7519}} and CBOR Web Tokens (CWTs) {{RFC8392}} as secure token formats, have vast possible applications. Some of these applications can involve issuing a token whereby certain semantics about the token can change over time, which are important to be able to communicate to relying parties in an interoperable manner, such as whether the token is considered invalidated or suspended by its issuer. A good example of such a case would be an "access token" as defined in section 1.4 of {{RFC6749}} where validity (after issuance, before expiration) is important meta information. A mechanism to gather this kind of information exists in the form of Token Introspection {{RFC7662}}, but requires direct interaction with the issuer, whereas this specification defines a mechanism using static information in the form of a status list.
```
paulbastian, line 49: The other use case for Issuer-Holder-Verifier is mentioned one paragraph further down. I think it makes sense to group them together

draft-ietf-oauth-status-list.md
```diff
@@ -48,6 +50,8 @@ JSON Web Tokens (JWTs) {{RFC7519}} and CBOR Web Tokens (CWTs) {{RFC8392}} as sec
...
 
+A good example for the usage of a status list would be an "access token" as defined in section 1.4 of {{RFC6749}} where validity (after issuance, before expiration) is important meta information. A mechanism to gather this kind of information exists in the form of Token Introspection {{RFC7662}}, but requires direct interaction with the issuer, whereas this specification defines a mechanism using static information in the form of a status list.
+
```
paulbastian, line 54:
```suggestion
This document defines Status List representations in JWT and CWT formats that describe the individual statuses of multiple Referenced Tokens, which themselves are also JWTs or CWTs. The statuses of all Referenced Tokens are conveyed via a bit array in the Status List. Each Referenced Token is allocated an index during issuance that represents its position within this bit array. The value of the bit(s) at this position correspond to the Referenced Token's status. The document also defines how an issuer of a Referenced Token references a Status List Token. Status Lists may be composed for expressing a range of Status Types. This document defines basic Status Types for the most common use cases as well as an extensibility mechanism for custom Status Types.

An example for the usage of a Status List would be an "access token" as defined in section 1.4 of {{RFC6749}} where validity (after issuance, before expiration) is important meta information. A mechanism to gather this kind of information exists in the form of Token Introspection {{RFC7662}}, but requires direct interaction with the issuer, whereas this specification defines a mechanism using static information in the form of a status list.

Another possible use case for the Status List is to express the status of verifiable credentials (Referenced Tokens) issued by an issuer in the Issuer-Holder-Verifier model.
```

paulbastian (approved):

draft-ietf-oauth-status-list.md
```diff
@@ -46,8 +48,11 @@ The status list data structures themselves are also represented as JWTs or CWTs.
...
 
+An example for the usage of a Status List would be an "access token" as defined in section 1.4 of {{RFC6749}} where validity (after issuance, before expiration) is important meta information. A mechanism to gather this kind of information exists in the form of Token Introspection {{RFC7662}}, but requires direct interaction with the issuer, whereas this specification defines a mechanism using static information in the form of a status list.
```
tplooker, line 53:
```suggestion
An example for the usage of a Status List is to manage the status of issued access tokens as defined in section 1.4 of {{RFC6749}}. Token Introspection {{RFC7662}} defines another way to determine the status of an issued access token, but it requires the party trying to validate an access tokens status to directly contact the token issuer, whereas the mechanism defined in this specification does not have this limitation.
```

PR #89, opened 2023-12-17
Restructure and Support for plain status list

paulbastian:
Closes #55
Closes #56
Closes #41

## 📑 Description
Adds support for unsigned Status List retrieved via HTTP endpoint.
Also restructures the document to allow this.

Remaining tasks:

- [x] Fix python example to provide Status List Object in 4.1
- [x] Fix python example in 5.1 to use the very same Status List Object from 4.1

## Preview Link

[click here for rendered preview of PR](https://vcstuff.github.io/draft-ietf-oauth-status-list/56-add-option-for-plain-status-list-unsigned/draft-ietf-oauth-status-list.html)

PR #90, opened 2023-12-20
Establish a mechanism that allows for other revocation means in the Referenced Token

### LINKED ISSUE #85: IETF 118: Establish a mechanism that allows for other revocation means in the Referenced Token
Leif asked to leave some space for future revocation mechanisms as StatusList might be only one approach.
The Structure in the Referenced Token should convey some kind of type, that enables other mechanisms reusing the "status" claim.

Related to #21 
Related to #40

c2bo:
Establish a mechanism that allows for other revocation means in the Referenced Token as discussed in #85. The only thing i changed was to call the reference "status_list" but I am open to change it to "status_bitstring" if we think that is a better fit.

Closes #85 

## Preview Link
[click here for rendered preview of PR](https://vcstuff.github.io/draft-ietf-oauth-status-list/c2bo/status-reference/draft-ietf-oauth-status-list.html)

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -97,7 +97,7 @@ The following rules apply to validating a JWT-based Status List Token. Applicati
 
 1. The JWT MUST contain an "iss" (issuer) claim that contains a unique string identifier for the entity that issued the JWT. In the absence of an application profile specifying otherwise, compliant applications MUST compare issuer values using the Simple String Comparison method defined in Section 6.2.1 of {{RFC3986}}. The value MUST be equal to that of the "iss" claim contained within the Referenced Token.
 
-2. The JWT MUST contain a "sub" (subject) claim that contains an unique string identifier for that Status List Token. The value MUST be equal to that of the "uri" claim contained in the "status" claim of the Referenced Token.
+2. The JWT MUST contain a "sub" (subject) claim that contains an unique string identifier for that Status List Token. The value MUST be equal to that of the "uri" claim contained in the "status_list" claim of the Referenced Token as defined in [](#jwt-referenced-token-status).
 
 3. The JWT MUST contain an "iat" (issued at) claim that identifies the time at which it was issued.
 
@@ -155,8 +155,10 @@ The following example is the decoded header and payload of a JWT meeting the pro
 {
   "iss": "https://example.com",
   "status": {
-    "idx": 0,
-    "uri": "https://example.com/statuslists/1"
+    "status_list": {
+      "idx": 0,
+      "uri": "https://example.com/statuslists/1"
+    }
   }
 }
 ~~~
@@ -167,9 +169,13 @@ The following rules apply to validating the "status" (status) claim
 
 1. The claim value MUST be a valid JSON object.
 
-2. The claim value object MUST contain an "idx" (index) member with a numeric value that represents the index to check for status information in the Status List for the current JWT. The value of this member MUST be a non-negative number, containing a value of zero or greater.
+2. The claim value object MUST contain a member called "status_list" that signals that status checks for this token can be done using the status list mechanism defined in this document. For the "status_list" object, the following rules apply:
 
-3. The claim value object MUST contain a "uri" member with a string value that identifies the Status List containing the status information for the JWT. The value of this member MUST be a uri conforming to {{RFC3986}}.
+    1. The claim value MUST be a valid JSON object.
+
+    2. The claim value object MUST contain an "idx" (index) member with a numeric value that represents the index to check for status information in the Status List for the current JWT. The value of this member MUST be a non-negative number, containing a value of zero or greater.
+
+    3. The claim value object MUST contain a "uri" member with a string value that identifies the Status List containing the status information for the JWT. The value of this member MUST be a uri conforming to {{RFC3986}}.
 
 # Status Types {#status-types}
 
@@ -467,6 +473,7 @@ for their valuable contributions, discussions and feedback to this specification
 
 -01
 
+* Change status claim to in referenced token to allow re-use for other mechanisms
 * Changing compression from gzip to zlib
 * Change typo in Status List Token sub claim description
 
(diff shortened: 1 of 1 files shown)


paulbastian (approved):
Looks good to me. Maybe merge after #89

paulbastian:
Should we consider registering an IANA registry?

draft-ietf-oauth-status-list.md
```diff
@@ -167,9 +169,13 @@ The following rules apply to validating the "status" (status) claim
 
 1. The claim value MUST be a valid JSON object.
 
-2. The claim value object MUST contain an "idx" (index) member with a numeric value that represents the index to check for status information in the Status List for the current JWT. The value of this member MUST be a non-negative number, containing a value of zero or greater.
+2. The claim value object MUST contain a member called "status_list" that signals that status checks for this token can be done using the status list mechanism defined in this document. For the "status_list" object, the following rules apply:
```
tplooker, line 172: If we make this a MUST then won't other future mechanisms only be additional rather than alternatives?

paulbastian: this is only a MUST for the mechanism used in the specification. If Giuseppe wants to implement an OCSP-like mechanism than it would look mlike:
```json
"status" : {
  "new-mechanism-name" : {
    "whatever-parameter" : "whatever-value"
  }
}  
```

paulbastian: I think we may need some text to the introduction that explains this in more detail. However its more better to merge #89 first and then start working from there

c2bo:
I'll add a small section as an introduction / to explain the `status` expansion idea (similar to `cnf` claim) over the weekend

draft-ietf-oauth-status-list.md
```diff
@@ -218,9 +218,10 @@ The Referenced Token MUST be encoded as a "JSON Web Token (JWT)" according to {{
...
-  * `idx`: REQUIRED. The `idx` (index) claim MUST specify an Integer that represents the index to check for status information in the Status List for the current Referenced Token. The value of `idx` MUST be a non-negative number, containing a value of zero or greater.
-  * `uri`: REQUIRED. The `uri` (URI) claim MUST specify a String value that identifies the Status List or Status List Token containing the status information for the Referenced Token. The value of `uri` MUST be a URI conforming to {{RFC3986}}.
+* `status`: REQUIRED. The `status` (status) claim MUST specify a JSON Object that contains a reference to a status mechanism.
+  * `status_list`: REQUIRED when the status list mechanism defined in this specification is used. It contains a reference to a Status List or Status List Token. The object contains exactly two claims:
+    * `idx`: REQUIRED. The `idx` (index) claim MUST specify an Integer that represents the index to check for status information in the Status List for the current Referenced Token. The value of `idx` MUST be a non-negative number, containing a value of zero or greater.
```
paulbastian, line 223: Maybe swap the order of Uri and idx?

paulbastian (commented):
Changes look good so far

tplooker (approved):

paulbastian:
I think the text in the introduction is a little weird. I checked what 7800 wrote:

> Other members of the "cnf"
   object may be defined because a proof-of-possession key may not be
   the only means of confirming the authenticity of the token

> This specification establishes the IANA "JWT Confirmation Methods"
   registry for these members in Section 6.2 and registers the members
   defined by this specification.  Other specifications can register
   other members used for confirmation, including other members for
   conveying proof-of-possession keys using different key representations.

I think the text may be more subtle along this lines. I would omit "we expect more in the future" which doesn't age well in a standard and be more neutral, just saying that we establish a registry for other mechanisms?

c2bo:
> I think the text in the introduction is a little weird. I checked what 7800 wrote:
> 
> > Other members of the "cnf"
> > object may be defined because a proof-of-possession key may not be
> > the only means of confirming the authenticity of the token
> 
> > This specification establishes the IANA "JWT Confirmation Methods"
> > registry for these members in Section 6.2 and registers the members
> > defined by this specification.  Other specifications can register
> > other members used for confirmation, including other members for
> > conveying proof-of-possession keys using different key representations.
> 
> I think the text may be more subtle along this lines. I would omit "we expect more in the future" which doesn't age well in a standard and be more neutral, just saying that we establish a registry for other mechanisms?

Sounds reasonable - i will propose a change

draft-ietf-oauth-status-list.md
c2bo: Current ordering is alphabetical which matches the examples and what afaik most json libraries will generate - that was my main intention for this ordering

c2bo:
I changed the text and realized that the IANA text in general needs some work, could you please re-review the current state?

draft-ietf-oauth-status-list.md
```diff
@@ -223,16 +227,21 @@ TBD
 
 # Referenced Token {#referenced-token}
 
+## Status Claim {#status-claim}
+
+By including a "status" claim in a JWT, the issuer of the JWT declares that the credential is referencing a mechanism to retrieve status information about this credential. The claim contains members used to reference to a status list as defined in this specification. Other members of the "status" object may be defined because status list means to check the status of a credential. This is analogous to "cnf" claim in Section 3.1 of {{RFC7800}} in which different authenticity confirmation methods can be included.
```
paulbastian, line 232: it seems this is in the wrong chapter?

paulbastian, line 232:
```suggestion
By including a "status" claim in a Referenced Token, the issuer is referencing a mechanism to retrieve status information about this Referenced Token. The claim contains members used to reference to a status list as defined in this specification. Other members of the "status" object may be defined by other specifications. This is analogous to "cnf" claim in Section 3.1 of {{RFC7800}} in which different authenticity confirmation methods can be included.
```
make it more general

paulbastian (changes requested):

draft-ietf-oauth-status-list.md
c2bo: I formulated this for JWT specifically as a starting point. This is similar to the JWT cnf registry. My current understanding is that we need to define 2 registries for "status" - one for JWT and one for CWT.
I guess I should adjust the title to match it - let's discuss in the editors call about this issue imho.

paulbastian (approved):

tplooker (approved):

Issue #121, opened 2024-03-13
TTL is missing for CWT/COSE examples


awoie:
The JWT spec has a ttl. The CWT/COSE examples should include that as well.

Issue #117, opened 2024-02-27
Register CWT ttl claim after CBOR/CWT PR is done


c2bo:
Register ttl claim for CWT

_Originally posted by @c2bo in https://github.com/vcstuff/draft-ietf-oauth-status-list/issues/101#issuecomment-1966222774_

PR #129, opened 2024-03-23
fix examples & adds ttl text for cwt

### LINKED ISSUE #127: Fix idx CBOR data type in examples
It appears the example uses a text string instead of a byte string for the idx value.
```
"idx": "0",
```

c2bo:
- fixes #127, #117
- adds ttl to the generated examples: #121 

Rendered verison: https://drafts.oauth.net/draft-ietf-oauth-status-list/c2bo/fix-examples/draft-ietf-oauth-status-list.html


draft-ietf-oauth-status-list.md
```diff
@@ -648,6 +649,11 @@ IANA "CBOR Web Token (CWT) Claims" registry {{IANA.CWT}} established by {{RFC839
 *  Change Controller: IETF
 *  Specification Document(s):  [](#status-list-token-cwt) of this specification
 
+* Claim Name: `ttl`
+* Claim Description: Time to Live
```
tplooker, line 653: As discussed on the call lets add the Claim Key entry to each IANA registration with the proposed values.

c2bo:
Should be fine now, with some minor fixes to rendering and adding the proposed values. Can you please check again @tplooker?

tplooker (approved):

PR #170, opened 2024-09-09
update security consideration

### LINKED ISSUE #73: Call for Adoption Feedback: Privacy concerns
https://mailarchive.ietf.org/arch/msg/oauth/hSOvzlKZTUJpI8ucB1CLNo5koqg/

> PS. A remainder:
> 
> - the protocol leaks information that allows verifiers to link the 
> tokens they receive.
> 
> -depending upon the architecture deployed by the token Issuer, the 
> Issuer may be in a position to act as Big Brother,
>            i.e. allowing it to know where and when a token it has issued 
> has been used.

### LINKED ISSUE #93: Consider adding an implementation consideration around using private relay style protocols for status list access
In order to mitigate issuers being able to track verifiers when retrieving a status list, consider adding an implementation consideration to recommend usage of a private relay protocol like [Oblivious HTTP](https://datatracker.ietf.org/doc/html/draft-ietf-ohai-ohttp).

### LINKED ISSUE #132: Privacy Consideration
Observability of Issuer's issuance behaviour, i.e. outsiders could observe how many tokens an issuer is issuing, therefore get information on his business.
Countermeasures: Maintain multiple lists in parallel, use decoy entries

### LINKED ISSUE #146: Implementation consideration to endianess
we currently assume there is no issue, but it may be worth to write it down

### LINKED ISSUE #165: Reference the JWT BCP

paulbastian:
Closes #146 
Closes #165 
Closes #132 
Closes #93 
Closes #73

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -35,6 +35,7 @@ normative:
   RFC7519: RFC7519
   RFC8259: RFC8259
   RFC8392: RFC8392
+  RFC8725: RFC8725
   RFC8949: RFC8949
   RFC9052: RFC9052
   RFC9110: RFC9110
@@ -571,7 +572,16 @@ Resulting in the byte array and compressed/base64url-encoded status list:
 # Security Considerations {#Security}
 
 ## Correct decoding and parsing of the encoded status list
-TODO elaborate on risks of incorrect parsing/decoding leading to erroneous status data
+
+Implementers should be particularly careful for the correct parsing and decoding of the status list. Incorrect implementations may check the index on the wrong data or misscalculate the bit and byte index leading to an erroneous status of the Referenced Token. Beware, that bits are indexed (bit order) from least signicant bit to most significant bit (also called "right to left") while bytes are indexed (byte order) in their natural incrementing byte order (usually written for display purpose from left to write). Endianess does not apply here because each status value fits within a single byte.
+
+Implementations shall always verify correctness using the test vectors given by this specification.
+
+## Security Guidance for JWT and CWT
+
+Status List Token in JWT format should follow the security considerations of {{RFC7519}} and the best current practices of {{RFC8725}}.
+
+Status List Token in CWT format should follow the security considerations of {{RFC8392}}.
 
 ## Cached and Stale status lists
 
@@ -579,12 +589,6 @@ When consumers or verifiers of the Status List fetch the data, they need to be a
 in the Status List Token provides one mechanism for setting a maximum cache time for the fetched data. This property permits distribution of
 a status list to a CDN or other distribution mechanism while giving guidance to consumers of the status list on how often they need to fetch
 a fresh copy of the status list even if that status list is not expired.
-
-## Authorized access to the Status List {#security-authorization}
-TODO elaborate on authorization mechanisms preventing misuse and profiling as described in privacy section
-
-## History
-TODO elaborate on status list only providing the up-to date/latest status, no historical data, may be provided by the underlying hosting architecture
 
 # Privacy Considerations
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -571,7 +572,16 @@ Resulting in the byte array and compressed/base64url-encoded status list:
 # Security Considerations {#Security}
 
 ## Correct decoding and parsing of the encoded status list
-TODO elaborate on risks of incorrect parsing/decoding leading to erroneous status data
+
+Implementers should be particularly careful for the correct parsing and decoding of the status list. Incorrect implementations may check the index on the wrong data or misscalculate the bit and byte index leading to an erroneous status of the Referenced Token. Beware, that bits are indexed (bit order) from least signicant bit to most significant bit (also called "right to left") while bytes are indexed (byte order) in their natural incrementing byte order (usually written for display purpose from left to write). Endianess does not apply here because each status value fits within a single byte.
```
c2bo, line 576:
```suggestion
Implementers should be particularly careful for the correct parsing and decoding of the status list. Incorrect implementations might check the index at the wrong position or miscalculate the bit and byte index leading to an erroneous status of the Referenced Token. Beware, that bits are indexed (bit order) from least significant bit to most significant bit (also called "right to left") while bytes are indexed (byte order) in their natural incrementing byte order (usually written for display purpose from left to write). Endianness does not apply here because each status value fits within a single byte.
```

paulbastian: with the first phrase "on the wrong data" I meant for example calculating index on the still compressed data

c2bo:
```suggestion
Implementers should be particularly careful for the correct parsing and decoding of the status list. Incorrect implementations might check the index at the wrong data or miscalculate the bit and byte index leading to an erroneous status of the Referenced Token. Beware, that bits are indexed (bit order) from least significant bit to most significant bit (also called "right to left") while bytes are indexed (byte order) in their natural incrementing byte order (usually written for display purpose from left to write). Endianness does not apply here because each status value fits within a single byte.
```

draft-ietf-oauth-status-list.md
```diff
@@ -571,7 +573,16 @@ Resulting in the byte array and compressed/base64url-encoded status list:
 # Security Considerations {#Security}
 
 ## Correct decoding and parsing of the encoded status list
-TODO elaborate on risks of incorrect parsing/decoding leading to erroneous status data
+
+Implementers should be particularly careful for the correct parsing and decoding of the status list. Incorrect implementations might check the index on the wrong data or miscalculate the bit and byte index leading to an erroneous status of the Referenced Token. Beware, that bits are indexed (bit order) from least significant bit to most significant bit (also called "right to left") while bytes are indexed (byte order) in their natural incrementing byte order (usually written for display purpose from left to write). Endianness does not apply here because each status value fits within a single byte.
+
+Implementations shall always verify correctness using the test vectors given by this specification.
```
c2bo, line 579: Should we use stronger language here? If yes, we have probably more parts in this section, where it would make sense.
```suggestion
Implementations are RECOMMENDED to verify correctness using the test vectors given by this specification.
```

c2bo (approved):
Not sure about the language in the security considerations section. Other RFCs seem to use normative language here - e.g., https://datatracker.ietf.org/doc/html/rfc9449#name-security-considerations.

Looks good apart from that part IMHO

draft-ietf-oauth-status-list.md
```diff
@@ -580,44 +591,52 @@ in the Status List Token provides one mechanism for setting a maximum cache time
...
 
-## Unobservability of Relying Parties {#privacy-relying-party}
+## Observability of Relying Parties {#privacy-relying-party}
 
 Once the Relying Party receives the Referenced Token, this enables him to request the Status List to validate its status through the provided `uri` parameter and look up the corresponding `index`. However, the Relying Party may persistently store the `uri` and `index` of the Referenced Token to request the Status List again at a later time. By doing so regularly, the Relying Party may create a profile of the Referenced Token's validity status. This behaviour may be intended as a feature, e.g. for a KYC process that requires regular validity checks, but might also be abused in cases where this is not intended and unknown to the Holder, e.g. profiling the suspension of a driving license or checking the employment status of an employee credential.
 
+TODO elaborate on status list only providing the up-to date/latest status, no historical data, may be provided by the underlying hosting architecture
```
tplooker, line 614: Are we leaving this in or shall we move it to an issue?

paulbastian: we had in before, so I would leave it in here

draft-ietf-oauth-status-list.md
```diff
@@ -612,7 +614,16 @@ Resulting in the byte array and compressed/base64url-encoded status list:
...
+
+Implementers should be particularly careful for the correct parsing and decoding of the status list. Incorrect implementations might check the index on the wrong data or miscalculate the bit and byte index leading to an erroneous status of the Referenced Token. Beware, that bits are indexed (bit order) from least significant bit to most significant bit (also called "right to left") while bytes are indexed (byte order) in their natural incrementing byte order (usually written for display purpose from left to write). Endianness does not apply here because each status value fits within a single byte.
+
+Implementations are RECOMMENDED to verify correctness using the test vectors given by this specification.
+
+## Security Guidance for JWT and CWT
+
+Status List Token in JWT format should follow the security considerations of {{RFC7519}} and the best current practices of {{RFC8725}}.
+
+Status List Token in CWT format should follow the security considerations of {{RFC8392}}.
```
tplooker, line 626:
```suggestion
A Status List Token in the CWT format should follow the security considerations of {{RFC8392}}.
```

paulbastian:
@tplooker I accepted your suggestions. You may approve and merge

tplooker (approved):

PR #172, opened 2024-09-09
add implementation consideration for Default Values and Double Alloca…

### LINKED ISSUE #24: Add implementations considerations
Add:
- prevent double allocating an index
- randomize allocation of indexes/avoid sequential usage

### LINKED ISSUE #47: requirement for status list size
As we declined to include a size property to the status list, we do not have other means than to decode&decompress the encoded status list and get status list size from the byte array length/size:
status list size = byte array size * 8 / bits
Therefore it makes to add the requirement that status list size must "fill up" to bytes

paulbastian:
Closes #47
Closes #24

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -642,6 +642,12 @@ The implementation is RECOMMENDED to initialize the Status List byte array with 
 
 The implementation is RECOMMENDED to prevent double allocation, i.e. re-using the same `uri` and `index` for mulitple Referenced Tokens. The Issuer MUST prevent any unintended double allocation by using the Status List.
 
+## Status List Size
+
+Status List Issuer may increase the size of a Status List if it requires indices for additional Referenced Tokens. It is RECOMMENDED that the size of a Status List fills up the byte boundary, i.e. `size` % `bits` = 0.
+
+The Status List Issuer may chunk its Referenced Tokens into multiple Status Lists to reduce the transmission size of an individual Status List Token. This may be useful for setups where some entities operate in constrained environments, e.g. for mobile internet or embedded devices.
+
 # IANA Considerations
 
 ## JSON Web Token Claims Registration
@@ -881,7 +887,7 @@ for their valuable contributions, discussions and feedback to this specification
 
 -04
 
-* add implementation consideration for Default Values and Double Allocation
+* add implementation consideration for Default Values, Double Allocation and Status List Size
 * add CORS considerations to the http endpoint
 * fix reference of Status List in CBOR format
 * added status_list CWT claim key assigned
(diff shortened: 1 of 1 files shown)


c2bo (approved):

draft-ietf-oauth-status-list.md
```diff
@@ -632,10 +632,22 @@ TODO evaluate definition of Status List Provider?
 
 The lifetime of a Status List (and the Status List Token) depends on the lifetime of its Referenced Tokens. Once all Referenced Tokens are expired, the Issuer may stop serving the Status List (and the Status List Token).
 
-Referenced Tokens may be regularly re-issued to increase security or to mitigate linkability and prevent tracking by Relying Parties. In this case, every Referenced Token MUST have a fresh Status List entry.
+Referenced Tokens may be regularly re-issued to increase security or to mitigate linkability and limit tracking by the Relying Parties. In this case, every re-issued Referenced Token MUST have a fresh Status List entry.
```
tplooker, line 635:
```suggestion
Referenced Tokens may be regularly re-issued to mitigate linkability of presentations to Relying Parties. In this case, every re-issued Referenced Token MUST have a fresh Status List entry in order to prevent this becoming possible source of correlation.
```

Minor nit

draft-ietf-oauth-status-list.md
```diff
@@ -632,10 +632,22 @@ TODO evaluate definition of Status List Provider?
...
 
+## Default Values and Double Allocation
+
+The implementation is RECOMMENDED to initialize the Status List byte array with a default value and provide this as an initialization parameter to the Issuer. The Issuer is RECOMMENDED to use a default value that represents the most common value for its Referenced Tokens to avoid an update during issuance.
+
+The implementation is RECOMMENDED to prevent double allocation, i.e. re-using the same `uri` and `index` for mulitple Referenced Tokens. The Issuer MUST prevent any unintended double allocation by using the Status List.
+
+## Status List Size
+
+Status List Issuer may increase the size of a Status List if it requires indices for additional Referenced Tokens. It is RECOMMENDED that the size of a Status List fills up the byte boundary, i.e. `size` % `bits` = 0.
```
tplooker, line 647:
```suggestion
The Status List Issuer may increase the size of a Status List if it requires indices for additional Referenced Tokens. It is RECOMMENDED that the size of a Status List in bits is divisible in bytes (8 bits) without a remainder, i.e. `size-in-bits` % 8 = 0.
```

paulbastian: I think we are talking about different points here. I'm talking about this issue: https://github.com/oauth-wg/draft-ietf-oauth-status-list/issues/47 which means the last byte should be filled and nto have some dangling bits

paulbastian: You are correct. I think yours is good as well

tplooker (approved):

PR #178, opened 2024-09-30
add optional support for historical status resolution

### LINKED ISSUE #138: Support an optional feature for historical resolution
We have a use case, in the financial field, to determine the time when the status of a given credential has changed. This requirement comes from US financial compliance regulation and is non negotiable.

The spec already supports an optional TTL which helps, but does not directly solve this issue. With a validity window (TTL), and consistent fetching (at the windows expiry) of status lists one could determine that a bit had flipped within a given window, and be able to approximate a timestamp of a given status change. Of course, this approach is noisy and inefficient.

To build off this section -- https://drafts.oauth.net/draft-ietf-oauth-status-list/draft-ietf-oauth-status-list.html#name-history -- we can make this type of query simpler. I would like to see language added to the spec that speaks to the (optional) capability of a status list server to support historical queries, or revisions of the document from a given time period. I believe this would maintain the privacy preserving characteristics of the specification while adding the ability for verifiers to approximate status change time, which is useful in a number of circumstances, especially as an optional-to-implement feature.

I am also open to other solutions that are similarly privacy preserving that support either directly or indirectly discerning 'time of status change', to date I have only seen solutions that would be far less privacy preserving, such as requesting specific instances (losing herd privacy).

c2bo:
Adds support for historical resolution of status as an optional feature.

Closes #138
Rendered Version: https://drafts.oauth.net/draft-ietf-oauth-status-list/c2bo/historical-resolution/draft-ietf-oauth-status-list.html

I will take another look at the wording / normative text and mark as ready for review when I am done.


draft-ietf-oauth-status-list.md
```diff
@@ -632,6 +650,34 @@ in the Status List Token provides one mechanism for setting a maximum cache time
...
 a fresh copy of the status list even if that status list is not expired.
 
+## Historical resolution
+By design, the Status List mechanism only conveys information about the state of a Referenced Token at the time the Status List Token was issued. The validity period for this information, as defined by the issuer, is explicitly stated by the `iat` (issued at) and `exp` (expiration time) claims for JWT, and their corresponding ones for the CWT representation.
+
+If support for historical status information needs to be supported, this can be achieved by extending the Status List Request as originally defined in [](#status-list-request) to support requesting specific timestamps. There are strong privacy concerns that have to be carefully taken into considerations when providing a mechanism that allows historic requests for status information - see [](#privacy-relying-party) for more details. Support for this functionality is purely optional and Implementers are RECOMMENDED to not support historic requests unless there are strong reasons to do so and after carefully considering the privacy implications.
```
paulbastian, line 656:
```suggestion
If support for historical status information is needed, this can be achieved by extending the request for the Status List as defined in [](#status-list-request) with a timestamp. There are strong privacy concerns that have to be carefully taken into considerations when providing a mechanism that allows historic requests for status information - see [](#privacy-relying-party) for more details. Support for this functionality is optional and Implementers are RECOMMENDED to not support historic requests unless there are strong reasons to do so and after carefully considering the privacy implications.
```

paulbastian (changes requested):

c2bo:
todo: 
- fix the references to this in privacy considerations
- fix some wording / language

decentralgabe (approved):
the language looks good, thanks for including it.

I will only add that while there may be privacy risks, there are legitimate business cases that require such historical data, such as guaranteeing compliance with financial regulation. In these cases, user privacy is made better by using this specification (as opposed to amore privacy-eroding solution), even if maximal privacy is not achieved. I hope the privacy section can speak to this nuance.

draft-ietf-oauth-status-list.md
```diff
@@ -755,6 +773,34 @@ in the Status List Token provides one mechanism for setting a maximum cache time
 a status list to a CDN or other distribution mechanism while giving guidance to consumers of the status list on how often they need to fetch
 a fresh copy of the status list even if that status list is not expired.
 
+## Historical resolution {#historical-resolution}
```
tplooker, line 776:
```suggestion
## Historical resolution {#historical-resolution}

```

draft-ietf-oauth-status-list.md
```diff
@@ -755,6 +773,34 @@ in the Status List Token provides one mechanism for setting a maximum cache time
...
+By design, the Status List mechanism only conveys information about the state of a Referenced Token at the time the Status List Token was issued. The validity period for this information, as defined by the issuer, is explicitly stated by the `iat` (issued at) and `exp` (expiration time) claims for JWT, and their corresponding ones for the CWT representation.
```
tplooker, line 777:
```suggestion
By design, the Status List mechanism only conveys information about the state of Reference Tokens at the time the Status List Token was issued. The validity period for this information, as defined by the issuer, is explicitly stated by the `iat` (issued at) and `exp` (expiration time) claims for JWT, and their corresponding ones for the CWT representation.
```

paulbastian (changes requested):
The whole text is in the Security Considerations section. I believe we should move the normative text to the section 8.

c2bo:
> The whole text is in the Security Considerations section. I believe we should move the normative text to the section 8.

Good poin - I moved the whole part to the end of Section 8.

draft-ietf-oauth-status-list.md
```diff
@@ -646,6 +664,35 @@ If this validation was not successful, the Referenced Token MUST be rejected. If
 
 If any of these checks fails, no statement about the status of the Referenced Token can be made and the Referenced Token SHOULD be rejected.
 
+## Historical resolution {#historical-resolution}
+
+By design, the Status List mechanism only conveys information about the state of Reference Tokens at the time the Status List Token was issued. The validity period for this information, as defined by the issuer, is explicitly stated by the `iat` (issued at) and `exp` (expiration time) claims for JWT, and their corresponding ones for the CWT representation.
```
paulbastian, line 669:
```suggestion
By default, the Status List mechanism only conveys information about the state of Reference Tokens at the time the Status List Token was issued. The validity period for this information, as defined by the issuer, is explicitly stated by the `iat` (issued at) and `exp` (expiration time) claims for JWT, and their corresponding ones for the CWT representation.
```

paulbastian (approved):

tplooker (approved):

PR #183, opened 2024-10-17
mixing formats and media types

### LINKED ISSUE #38: Mixing JSON and CBOR formats
Is it allowed/recommended to have Status List Token in one format and the Referenced Token in another format?

### LINKED ISSUE #137: Set content type to optional than must
Some CDNs like github repos/pages only provide static hosting and therefore no way to set the content-type to another value.

For the JOSE case I could publish a txt or json file.
So when it's a json file, it should be equal to `application/statuslist+json`, in case of a JWT it is `application/statuslist+jwt`.

### LINKED ISSUE #168: Support for content negotiation as denoted in the standard is limited for some CDNs and http servers
As described in https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-03.html#section-8.1 content negotiation using the accept header is used when requesting a status list / status list token. This is used to request different formats depending on the needs of the caller using a single URI for the status list.

This mechanism in general is valid and follows the http standards. Implementing this works well when an application server is used to serve the status lists.

Nevertheless in practice this causes problems in some scenarios. Several CDNs and services that provide caches of other HTTP resources do not implement the required functionality correctly.

For example [cloudflare, does explicitly ignore the accept header even if the vary http response header is used](https://developers.cloudflare.com/cache/concepts/cache-control/#other). Some other services offer this functionality tough, for example Amazon CloudFront.

In addition nginx can not easily handle content negotiation using the accept header correctly. Using the a map section together with the Accept header allows to choose different resources based on the requested content type. This is only working for single value Accept headers though. More complex header values as `Accept: application/statuslist+*;q=0.5,application/statuslist+cwt` can not easily be handled. The current state of the standard does not mention such option but does not forbid them either. Apache2 has https://httpd.apache.org/docs/current/content-negotiation.html, that could solve this. I did not evaluate that further.

So even though the mechanism is good and simple as is and follows http standards, it might be difficult to implement when using a CDN to scale serving the status list.

A possible solution would be to use distinct URIs to request the status list depending on the type (for example by appending extensions like jwt, cwt, json and cbor) instead of the request header. This seems to me as a workaround and a suboptimal solution though. As the specification as it is is totally valid IMHO it could as well been left as it is, which means the need for implementation of this functionality in the used components will arise and the usage of certain CDNs is not possible.

Related to #137 but addresses the request handling instead of response generation.

c2bo:
closes #38
closes #168
closes #137

Rendered preview: https://drafts.oauth.net/draft-ietf-oauth-status-list/c2bo/mixing-formats/draft-ietf-oauth-status-list.html

This PR attempts to add a bit of language about mixing formats and also changes the content-type in the status list response to SHOULD and language around content type parsing that is similar to the language used in HTTP/1.1.

I am happy to remove the media types part if we want a bit more discussion, but when I started writing this, it felt like right path so I decided to include it as a proposal.

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -835,14 +835,13 @@ The Status List Issuer may chunk its Referenced Tokens into multiple Status List
 
 ## Status List Formats
 
- This specification defines 4 different formats that can be used to provide a status list:
+ This specification defines 4 different formats of the status list:
 
  - JSON
  - JWT
  - CBOR
  - CWT
 
-This decision was made to allow the highest degree of freedom to use the variant that best fits the ecosystem & use-case intending to add a status or validity information.
 
 This specification states no requirements to not mix different formats like a CBOR based token using a JWT for the status list, but the expectation is that within an ecosystem, a choice for specific formats is made.
 Within such an ecosystem, only support for those selected variants is required and implementations should know what to expect via a profile.
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -1079,6 +1079,11 @@ for their valuable contributions, discussions and feedback to this specification
 # Document History
 {:numbered="false"}
 
+
+-05
+
+* add section about mixing status list formats and media type
+
 -04
 
 * add mDL example as Referenced Token and consolidate CWT and CBOR sections
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -843,7 +843,7 @@ The Status List Issuer may chunk its Referenced Tokens into multiple Status List
  - CWT
 
 
-This specification states no requirements to not mix different formats like a CBOR based token using a JWT for the status list, but the expectation is that within an ecosystem, a choice for specific formats is made.
+This specification states no requirements to not mix different formats like a CBOR based Referenced Token using a JWT for the Status List, but the expectation is that within an ecosystem, a choice for specific formats is made.
 Within such an ecosystem, only support for those selected variants is required and implementations should know what to expect via a profile.
 
 The return type of a status list should be signaled by media type as described in [](#status-types-response). If no media type was provided and the expected type is not clearly defined the ecosystem, implementations should be robust to parsing errors. In that case, a client MAY attempt to guess the media type via inspection of the response or name extension(s) of the URI.
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -847,7 +847,7 @@ This decision was made to allow the highest degree of freedom to use the variant
 This specification states no requirements to not mix different formats like a CBOR based token using a JWT for the status list, but the expectation is that within an ecosystem, a choice for specific formats is made.
 Within such an ecosystem, only support for those selected variants is required and implementations should know what to expect via a profile.
 
-The return type of a status list should be signaled by media type as described in [](#status-types-response). If no media type was provided and the expected type is not clearly defined the ecosystem, implementations should be robust to parsing errors. In that case, a client MAY attempt to guess the media type via inspection of the response or via name extension(s) of the URI.
+The return type of a status list should be signaled by media type as described in [](#status-types-response). If no media type was provided and the expected type is not clearly defined the ecosystem, implementations should be robust to parsing errors. In that case, a client MAY attempt to guess the media type via inspection of the response or name extension(s) of the URI.
 
 # IANA Considerations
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -833,6 +833,22 @@ The Status List Issuer may increase the size of a Status List if it requires ind
...
+## Status List Formats
+
+ This specification defines 4 different formats that can be used to provide a status list:
+
+ - JSON
+ - JWT
+ - CBOR
+ - CWT
+
+This decision was made to allow the highest degree of freedom to use the variant that best fits the ecosystem & use-case intending to add a status or validity information.
```
tplooker, line 845:
```suggestion
```

Generally we shouldn't document decisions in spec text.

draft-ietf-oauth-status-list.md
```diff
@@ -833,6 +833,22 @@ The Status List Issuer may increase the size of a Status List if it requires ind
 
 The Status List Issuer may chunk its Referenced Tokens into multiple Status Lists to reduce the transmission size of an individual Status List Token. This may be useful for setups where some entities operate in constrained environments, e.g. for mobile internet or embedded devices.
 
+## Status List Formats
+
+ This specification defines 4 different formats that can be used to provide a status list:
```
tplooker, line 838:
```suggestion
 This specification defines 4 different formats of the status list:
```

draft-ietf-oauth-status-list.md
```diff
@@ -607,9 +607,9 @@ The Relying Party SHOULD send the following Accept-Header to indicate the reques
 
 If the Relying Party does not send an Accept Header, the response type is assumed to be known implicit or out-of-band.
 
-## Status List Response
+## Status List Response {#status-types-response}
```
paulbastian, line 610: Why is the reference called status-**types**-response?

draft-ietf-oauth-status-list.md
```diff
@@ -833,6 +833,21 @@ The Status List Issuer may increase the size of a Status List if it requires ind
...
+
+ This specification defines 4 different formats of the status list:
+
+ - JSON
+ - JWT
+ - CBOR
+ - CWT
+
+
+This specification states no requirements to not mix different formats like a CBOR based token using a JWT for the status list, but the expectation is that within an ecosystem, a choice for specific formats is made.
```
paulbastian, line 846:
```suggestion
This specification states no requirements to not mix different formats like a CBOR based Referenced Token using a JWT for the Status List, but the expectation is that within an ecosystem, a choice for specific formats is made.
```

paulbastian (changes requested):
I'm not feeling confident with the change of MUST -> SHOULD for media type return. Maybe "MUST .. unless media type is known out-of-band"?

draft-ietf-oauth-status-list.md
c2bo: copy paste mistake - fixed it

draft-ietf-oauth-status-list.md
```diff
@@ -607,9 +607,9 @@ The Relying Party SHOULD send the following Accept-Header to indicate the reques
 
 If the Relying Party does not send an Accept Header, the response type is assumed to be known implicit or out-of-band.
 
-## Status List Response
+## Status List Response {#status-list-response}
 
-In the successful response, the Status List Provider MUST use the following content-type:
+In the successful response, the Status List Provider SHOULD use the following content-type:
```
c2bo, line 612: Would this work for you @paulbastian?
```suggestion
In the successful response, the Status List Provider MUST use the following content-type unless known out of band:
```

draft-ietf-oauth-status-list.md
```diff
@@ -916,6 +916,21 @@ The Status List Issuer may increase the size of a Status List if it requires ind
...
+ - JSON
+ - JWT
+ - CBOR
+ - CWT
+
+
+This specification states no requirements to not mix different formats like a CBOR based Referenced Token using a JWT for the Status List, but the expectation is that within an ecosystem, a choice for specific formats is made.
+Within such an ecosystem, only support for those selected variants is required and implementations should know what to expect via a profile.
+
+The return type of a status list should be signaled by media type as described in [](#status-list-response). If no media type was provided and the expected type is not clearly defined the ecosystem, implementations should be robust to parsing errors. In that case, a client MAY attempt to guess the media type via inspection of the response or name extension(s) of the URI.
```
c2bo, line 932:
```suggestion
```

draft-ietf-oauth-status-list.md
```diff
@@ -655,9 +655,9 @@ Content-Type: application/statuslist+jwt
 {::include ./examples/status_list_jwt_raw}
 ~~~
 
-## Status List Response
+## Status List Response {#status-list-response}
 
-In the successful response, the Status List Provider MUST use the following content-type:
+In the successful response, the Status List Provider MUST use the following content-type unless known out of band:
```
c2bo, line 660:
```suggestion
In the successful response, the Status List Provider MUST use the following content-type:
```

PR #185, opened 2024-10-20
remove constraint that Status List Token must not use a MAC

### LINKED ISSUE #111: Relax constraint on not allowing MAC for Status List Token?
As the similar requirement was relaxed on SD-JWT, I think we should consider the same.

paulbastian:
Closes #111

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -302,7 +302,7 @@ The following additional rules apply:
 
 1. The CWT MAY contain other claims.
 
-2. The CWT MUST be secured using a cryptographic signature or MAC algorithm. Relying Parties MUST reject JWTs with an invalid signature.
+2. The CWT MUST be secured using a cryptographic signature or MAC algorithm. Relying Parties MUST reject CWTs with an invalid signature.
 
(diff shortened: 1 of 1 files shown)


c2bo (changes requested):
Document history entry is missing and I think we should also adjust line 90

> A Status List may either be provided via HTTPS or be signed and embedded into a Status List Token, whereas this document defines its representations in JWT and CWT.

c2bo (approved):

draft-ietf-oauth-status-list.md
```diff
@@ -302,11 +302,11 @@ The following additional rules apply:
 
 1. The CWT MAY contain other claims.
 
-2. The CWT MUST be digitally signed using an asymmetric cryptographic algorithm. Relying parties MUST reject the CWT if it is using a Message Authentication Code (MAC) algorithm. Relying parties MUST reject CWTs with an invalid signature.
+2. The CWT MUST be secured using a cryptographic signature or MAC algorithm. Relying Parties MUST reject JWTs with an invalid signature.
```
tplooker, line 305:
```suggestion
2. The CWT MUST be secured using a cryptographic signature or MAC algorithm. Relying Parties MUST reject CWTs with an invalid signature.
```

tplooker (commented):
One typo otherwise looks good.

tplooker (approved):

PR #186, opened 2024-10-20
improve section on Status Types and introduce IANA registry for it

### LINKED ISSUE #177: Missing IANA section for status types
The section 7 explains how to setup the registry (https://drafts.oauth.net/draft-ietf-oauth-status-list/draft-ietf-oauth-status-list.html#name-status-types) but we are missing a corresponding section in the IANA appendix

### LINKED ISSUE #181: Add Reserved For Private Use Range
The first few entries are valid, revoked, suspended.

There may be future well defined values that issuers and verifiers want global interoperability around.

There may also be a need for private use of this space.

This document should describe the "reserved for private use range", and leave room for enough globally interoperable and registered status values

paulbastian:
Closes #177 
Closes #181

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -570,7 +570,7 @@ The following is the CBOR Diagnostic Notation of the example above:
 
 # Status Types {#status-types}
 
-This document defines statuses of Referenced Tokens as Status Type values. A status describes the state, mode, condition or stage of an entity that represented ny the Referenced Token.
+This document defines statuses of Referenced Tokens as Status Type values. A status describes the state, mode, condition or stage of an entity that is represented by the Referenced Token.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -570,23 +570,20 @@ The following is the CBOR Diagnostic Notation of the example above:
 
 # Status Types {#status-types}
 
-This document defines potential statuses of Referenced Tokens as Status Type values. If the Status List contains more than one bit per token (as defined by "bits" in the Status List), then the whole value of bits MUST describe one value. A Status List can not represent multiple statuses per Referenced Token.
+This document defines statuses of Referenced Tokens as Status Type values. A status describes the state, mode, condition or stage of an entity that represented ny the Referenced Token.
```
tplooker, line 573:
```suggestion
This document defines statuses of Referenced Tokens as Status Type values. A status describes the state, mode, condition or stage of an entity that is represented by the Referenced Token.
```

Is this what you meant?

c2bo (approved):

tplooker (approved):

PR #201, opened 2024-11-24
status codes & redirect

### LINKED ISSUE #188: Define Status List Response status code & redirection behavior
We do not define how to respond to a status list request but have status codes on examples
We might also need to add language for redirects?

c2bo:
Closes #188

Rendered version: https://drafts.oauth.net/draft-ietf-oauth-status-list/c2bo/status-code-redirect/draft-ietf-oauth-status-list.html

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -636,7 +636,7 @@ If the Relying Party does not send an Accept Header, the response type is assume
 
 A successful response that contains a Status List Token MUST use an HTTP status code in the 2xx range. If the Status List Token was not found, the response MUST use the HTTP status code 404.
 
-A response MAY also redirect the client to another URI using a HTTP status code in the 3xx range, which clients SHOULD follow. A client SHOULD detect and intervene in cyclical redirections (i.e., "infinite" redirection loops).
+A response MAY also choose to redirect the client to another URI using a HTTP status code in the 3xx range, which clients SHOULD follow. A client SHOULD detect and intervene in cyclical redirections (i.e., "infinite" redirection loops).
 
 The following are non-normative examples for a request and response for a Status List Token with type `application/statuslist+jwt`:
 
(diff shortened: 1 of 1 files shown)


paulbastian (approved):

draft-ietf-oauth-status-list.md
```diff
@@ -634,6 +634,10 @@ The Relying Party SHOULD send the following Accept-Header to indicate the reques
 
 If the Relying Party does not send an Accept Header, the response type is assumed to be known implicit or out-of-band.
 
+A successful response that contains a Status List Token MUST use an HTTP status code in the 2xx range. If the Status List Token was not found, the response MUST use the HTTP status code 404.
+
+A response MAY also redirect the client to another URI using a HTTP status code in the 3xx range, which clients SHOULD follow. A client SHOULD detect and intervene in cyclical redirections (i.e., "infinite" redirection loops).
```
paulbastian, line 639: is redirect rather a SHOULD?

c2bo, line 639:
```suggestion
A response MAY also choose to redirect the client to another URI using a HTTP status code in the 3xx range, which clients SHOULD follow. A client SHOULD detect and intervene in cyclical redirections (i.e., "infinite" redirection loops).
```

draft-ietf-oauth-status-list.md
```diff
@@ -634,6 +634,10 @@ The Relying Party SHOULD send the following Accept-Header to indicate the reques
 
 If the Relying Party does not send an Accept Header, the response type is assumed to be known implicit or out-of-band.
 
+A successful response that contains a Status List Token MUST use an HTTP status code in the 2xx range. If the Status List Token was not found, the response MUST use the HTTP status code 404.
```
tplooker, line 637:
```suggestion
A successful response that contains a Status List Token MUST use an HTTP status code in the 2xx range.
```

While I think your right with regard to the majority of implementations here I dont think it pays to call out just 404, otherwise why not other 4xx range codes and or 5xx range codes too?

tplooker (approved):
Ok to merge with the minor suggestion.

PR #204, opened 2024-11-29
update security considerations

### LINKED ISSUE #199: Re-work Security considerations
- Only cryptographically protected variants are considered --> adjust security considerations accordingly if necessary

c2bo:
Closes #199

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -803,6 +803,8 @@ Resulting in the byte array and compressed/base64url-encoded Status List:
 
 # Security Considerations {#Security}
 
+The Status List as defined in this specification only exists in cryptographically secured containers which allows checking the integrity and origin without relying on other aspects like transport security (e.g., the web PKI).
+
 ## Correct decoding and parsing of the encoded Status List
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -818,6 +818,8 @@ Resulting in the byte array and compressed/base64url-encoded Status List:
 
 # Security Considerations {#Security}
 
+The Status List as defined in this specification only exists in cryptographically secured containers which allows checking the integrity and origin without relying on other aspects like transport security (e.g., the web PKI).
```
paulbastian, line 821: this is a little confusing to me, are we talking about the Status List object, that is only communciated in Status List Tokens or about referenced Tokens?

c2bo: status list object currently (after the change) only exists in secured containers

paulbastian: should we add reference to the section?

paulbastian (approved):

draft-ietf-oauth-status-list.md
c2bo:
```suggestion
The Status List as defined in [](#status-list) only exists in cryptographically secured containers which allows checking the integrity and origin without relying on other aspects like transport security (e.g., the web PKI).
```

tplooker (approved):

PR #206, opened 2024-12-02
additional text for iana registry

### LINKED ISSUE #194: IANA Registry
- Section 4 of RFC 8126 -> supply registration procedures for the registries
- Define where these registries should be placed

c2bo:
Closes #194 

We would register:
CWT Status Mechanisms registry in CWT 
JWT Status Mechanisms registry in JWT 
Status Types registry in OAuth parameters

Any ideas how we could simplify this?

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -968,7 +968,7 @@ IANA "JSON Web Token Claims" registry {{IANA.JWT}} established by {{RFC7519}}.
 
 This specification establishes the IANA "Status Mechanism Methods" registry for JWT "status" member values. The registry records the status mechanism method member and a reference to the specification that defines it.
 
-JWT Status Mechanism Methods are registered by Specification Required [RFC5226] after a two-week
+JWT Status Mechanism Methods are registered by Specification Required [RFC5226] after a three-week
 review period on the jwt-reg-review@ietf.org mailing list, on the advice of one or more Designated Experts.
 However, to allow for the allocation of names prior to publication, the Designated Expert(s) may approve
 registration once they are satisfied that such a specification will be published.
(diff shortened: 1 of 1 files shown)


paulbastian:
looks good to me in general. I would extend this to more than two weeks, if I look back how the current year went by. I think this is a good start for now, so we should merge this first and then evaluate afterwards if we can merge both status registries.

draft-ietf-oauth-status-list.md
```diff
@@ -968,6 +968,19 @@ IANA "JSON Web Token Claims" registry {{IANA.JWT}} established by {{RFC7519}}.
 
 This specification establishes the IANA "Status Mechanism Methods" registry for JWT "status" member values. The registry records the status mechanism method member and a reference to the specification that defines it.
 
+JWT Status Mechanism Methods are registered by Specification Required [RFC5226] after a two-week
```
c2bo, line 971:
```suggestion
JWT Status Mechanism Methods are registered by Specification Required [RFC5226] after a three-week
```

draft-ietf-oauth-status-list.md
```diff
@@ -1034,6 +1047,18 @@ IANA "CBOR Web Token (CWT) Claims" registry {{IANA.CWT}} established by {{RFC839
 
 This specification establishes the IANA "Status Mechanism Methods" registry for CWT "status" member values. The registry records the status mechanism method member and a reference to the specification that defines it.
 
+CWT Status Mechanism Methods are registered by Specification Required [RFC5226] after a two-week
```
c2bo, line 1050:
```suggestion
CWT Status Mechanism Methods are registered by Specification Required [RFC5226] after a three-week
```

c2bo:
> looks good to me in general. I would extend this to more than two weeks, if I look back how the current year went by. I think this is a good start for now, so we should merge this first and then evaluate afterwards if we can merge both status registries.

JWT/CWT registries use three week periods, the OAuth Parameter one uses two weeks. Should we just align with these registries for each one?

paulbastian (approved):

tplooker (approved):

PR #211, opened 2024-12-15
differentiate unlinkability between Issuer-RP and RP-RP

### LINKED ISSUE #207: Proposed rewording of section 12.5.  Unlinkability
It is proposed to include another section called : 

> 12.5.2  Alternative mechanism to the use of Token Status Lists

Before it, a section that will re-use the existing text will be called :

> 12.5.1  Further recommendations to the use of Token Status Lists

The whole text would look as follows:

**12.5.1  Further recommendations to the use of Token Status Lists**

   To avoid privacy risks for colluding Relying Parties, it is RECOMMENDED 
   that Issuers use batch issuance to issue multiple tokens, see Section 13.1.

   To avoid further correlatable information by the values of uri and
   index, Issuers are RECOMMENDED to:

   *  choose non-sequential, pseudo-random or random indices

   *  use decoy entries to obfuscate the real number of Referenced
      Tokens within a Status List

   *  choose to deploy and utilize multiple Status Lists simultaneously

**12.5.2  Alternative mechanism to the use of Token Status Lists**

   To avoid privacy risks for colluding Relying Parties, Issuers SHOULD 
   use batch issuance of multiple Tokens, but it is possible to use a 
   mechanism different from the use of Token Status Lists.   In such a 
   case, the cryptographically secured data structure derived from the 
   Token by the Holder does not contain a reference to a Token Status 
   List. 
      
   This mechanism is only usable if the Issuer has been able to verify, 
   at the time of the issuance of a Token, that an appropriate application
   running under a secure environment is supported by the Holder.

   Instead of using a protocol for fetching token status lists from a 
   Status Issuer, it is possible to take advantage of another protocol 
   between an Holder and an Issuer, when that protocol is supported.

   The fact that this other protocol is supported by an appropriate 
   application running under a secure environment can be indicated by 
   the Issuer to the Verifier by including into the Token, for example, 
   a security policy similar to the certificate policy from X.509 certificates 
   (RFC 5280). 
   See: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.4

   When the Relying Party is able to know that this protocol between 
   an Holder and an Issuer is supported, it knows that the Issuer will 
   be able, when necessary, to either suspend or invalidate any Token 
   that is has issued. 
    
   The protocol works as follows: 

   from time to time, the Holder queries each Issuer that has issued 
   a Token, and if a Token that it has issued needs to be suspended 
   or invalidated, the Issuer informs the Holder.   In that case, the 
   Holder will  immediately stop creating new cryptographically 
   secured data structures derived from suspended or invalidated 
   Tokens.

   When a Token has been invalidated, the process cannot be reversed. 
   However, when a Token has been suspended, the process can be reversed.

   If the Holder has been unable to contact an Issuer during a defined 
   period of time, e.g., 24 hours, the Holder will stop creating new 
   cryptographically secured data structures derived from Tokens issued
   by that Issuer.

   In this way, as long as the Holder is able to connect to each of the 
   Issuers from time to time, this mechanism can be used in an off-line 
   environment. 

   In order to check the suspension or the revocation status of a 
   cryptographically secured data structure, neither the Holder nor the 
   Verifier needs to fetch additional information.  When digital 
   credentials are used, the Holder and the Verifier can interact in a 
   proximity mode without using any on-line protocol.

### LINKED ISSUE #209: Revocation and suspension are not that complicated and there are easy tradeoffs
In #207, Paul wrote:

> Revocation is complicated and there are no easy tradeoffs, we shouldn't let the reader believe this.

What you write was true before smart phones existed, but the existence of smart phones offers new possibilities. 
If / when trusted apps. can be used on a smart phone, then other status methods become possible.

**Revocation and suspension of digital credentials**

When a Holder is requesting a digital credential, it needs to know the url of the Issuer, while the Issuer needs to know the characteristics of the Holder, e.g., using attestations. This allows to establish a secure channel between them. That channel can then be used to support a "status protocol" between a Holder and an Issuer.

It is important to notice that this alternative method can support both SUSPENSION and revocation. 
As soon as either a suspension or a revocation is decided by the Issuer, it can be taken into account either very quickly by the app. if the smartphone is connected to the Internet, e.g. within five minutes when the app. polls each Issuer every five minutes, or within 24 hours, if the smartphone is not connected to the Internet.

In addition, this alternative method natively supports the issuer-verifier unlinkability characteristic. 

In terms of energy saving, it is "green", since no data centre needs to be used, e.g., to make available large numbers of Token Status Lists.

Up to now, I didn't proposed a name to identify this alternative method. I propose: **status_protocol**

   *  Status Method Value: status_protocol

   *  Status Method Description: A protocol between a Holder and an Issuer where the Issuer instructs the Holder to revoke, 
      suspend or reactivate one or more tokens.

To close the loop, the **status_protocol** method can be considered to be an easy tradeoff. 

:-)

More seriously, it shifts the problem into two other areas: 
-	the downloading of wallet applications into smart phones by trusted wallets providers, and
-	the development of a status protocol between an Holder and an Issuer. 
        In that case, the status protocol could be defined in a standard track RFC.

R**evocation and suspension at once of all the digital credentials contained in a lost or stolen digital identity wallet**

A similar approach can be used to address the case where the Holder is lost or stolen (in practice, when the smart phone is lost or stolen). Instead of placing the burden on the user to contact individually each Issuer that issued a digital credential, it is easier for the user to use a single point of contact: the Holder Provider (i.e., the app. provider).

An app. on a smart phone, already regularly connects to its provider, e.g., to check for software updates. 
An additional protocol to revoke, suspend or reactivate all the digital credentials / tokens hosted by the Holder would be added.

paulbastian:
Closes #207 
Closes #209

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -911,7 +911,7 @@ The tuple of uri and index inside the Referenced Token are unique and therefore 
 
 ### Colluding Relying Parties
 
-Two or more colluding Relying Parties may link two transaction involving the same Referenced Tokens by comparing the status claims of received Referenced Tokens, and therefore determine that they have interacted with the same Holder.
+Two or more colluding Relying Parties may link two transactions involving the same Referenced Token by comparing the status claims of received Referenced Tokens, and therefore determine that they have interacted with the same Holder.
 
 To avoid privacy risks for colluding Relying Parties, it is RECOMMENDED that Issuers use batch issuance to issue multiple Referenced Tokens, see [](#implementation-lifecycle). To avoid further correlatable information by the values of `uri` and `index`, Status Issuers are RECOMMENDED to:
 
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -152,8 +152,6 @@ Revocation mechanisms are an essential part for most identity ecosystems. In the
 
 This specification seeks to find a balance between scalability, security, and privacy by minimizing the status information to mere bits (often a single bit) and compressing the resulting binary data. Thereby, a Status List may contain statuses of many thousands or millions Referenced Tokens while remaining as small as possible. Placing large amounts of Referenced Tokens into the same list also enables herd privacy relative to the Status Provider.
 
-This specification establishes the IANA "Status Mechanism Methods" registry for status mechanism and registers the members defined by this specification. Other specifications can register other members used for status retrieval.
-
 ## Design Considerations
 
 The decisions taken in this specification aim to achieve the following design goals:
@@ -166,6 +164,10 @@ The decisions taken in this specification aim to achieve the following design go
 * the specification shall support JSON and CBOR based tokens
 * the specification shall not specify key resolution or trust frameworks
 * the specification shall design an extension point to convey information about the status of a token that can be re-used by other mechanisms
+
+## Status Mechanism Registry
+
+This specification establishes the IANA "Status Mechanism Methods" registry for status mechanism and registers the members defined by this specification. Other specifications can register other members used for status retrieval. Other status mechanisms may have different tradeoffs regarding security, privacy, scalability adn complexity. The privacy and security considerations in this document only represent the properties of the Status List mechanism.
 
 # Conventions and Definitions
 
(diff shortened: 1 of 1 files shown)


paulbastian:
I've seen that we currently say:

> IANA " JWT Status Mechanism Methods" registry ...

I was wondering if we should cut out "Methods"?

draft-ietf-oauth-status-list.md
```diff
@@ -905,16 +907,22 @@ This behaviour could be mitigated by:
 
 ## Unlinkability
 
-Colluding Issuers and Relying Parties have the possibility to link two transactions, as the tuple of `uri` and `index` inside the Referenced Token are unique and therefore traceable data. By comparing the status claims of received Referenced Tokens, two colluding Relying Parties could determine that they have interacted with the same user or an Issuer could trace the usage of its issued Referenced Token by colluding with various Relying Parties. It is therefore recommended to use Status Lists for Referenced Token formats that have similar unlinkability properties.
+The tuple of uri and index inside the Referenced Token are unique and therefore is traceable data.
+
+### Colluding Relying Parties
 
-To avoid privacy risks for colluding Relying Parties, it is RECOMMENDED that Issuers use batch issuance to issue multiple tokens, see [](#implementation-lifecycle).
+Two or more colluding Relying Parties may link two transaction involving the same Referenced Tokens by comparing the status claims of received Referenced Tokens, and therefore determine that they have interacted with the same Holder.
```
c2bo, line 914:
```suggestion
Two or more colluding Relying Parties may link two transactions involving the same Referenced Token by comparing the status claims of received Referenced Tokens, and therefore determine that they have interacted with the same Holder.
```

c2bo:
> I've seen that we currently say:
> 
> > IANA " JWT Status Mechanism Methods" registry ...
> 
> I was wondering if we should cut out "Methods"?

Hmm, Yeah I guess it is somewhat redundant - would be fine for me

tplooker:
> Hmm, Yeah I guess it is somewhat redundant - would be fine for me

Yep I agree "Methods" in this context is redundant, so we should be able to drop.

tplooker (approved):

PR #213, opened 2025-01-02
Nit - change "write" to "right" for describing directionality

dhs-BI:
Minor change to address the use of the homonym "write" instead of "right".


c2bo (approved):
Good catch, thank you!

paulbastian (approved):

tplooker (approved):

PR #214, opened 2025-01-06
rename Status Mechanism Methods registry to Status Mechanisms registry

paulbastian:


c2bo:
There is still a bunch of mentions of `method` after the changed parts. IMHO, let's change everywhere?

c2bo (approved):

paulbastian:
> There is still a bunch of mentions of `method` after the changed parts. IMHO, let's change everywhere?

Thanks, I only searched for "Methods" and forgot about "Method", should be fixed now

tplooker (approved):

PR #248, opened 2025-01-30
add recommendations for Key Resolution and Trust Management

### LINKED ISSUE #224: Interims Feedback: Explain motivation to split issuer / status list issuer / status list provider

### LINKED ISSUE #227: Which keys should be used to sign and verify Status List Tokens ?
In section 5 (Status List Token) the text states:

> A Status List Token embeds the Status List into a token that is cryptographically signed and protects the integrity of the Status List.

This sentence does not state which key should be used to sign or to verity the Status List Token.

Using JOSE, the "iss" (issuer) claim identifies the principal that issued the JWT. The iss" claim is not present in the list of claims from section 5.1 (Status List Token in JWT Format).

Section 8.3 which is about "Validation Rules" states:

>    3.  Validate the Status List Token:
> 
>        1.  Validate the Status List Token by following the rules defined
>            in section 7.2 of [RFC7519] for JWTs

However, in section 7.2 from RFC 7519, the single "occurrence of "iss" is in the following sentence:

> Some applications may include case-insensitive information in a case-sensitive value, such as including a DNS name as part of the "iss" (issuer) claim value.

The text does not say that the content of "iss" claim shall be used and does not specific against which value it shall match.

Let us make a parallel with OCSP, section 4.2.2.2 from RFC 6960 which states:

> 4.2.2.2.  Authorized Responders
> The key that signs a certificate's status information need not be the same key that signed the certificate.  It is necessary, however, to ensure that the entity signing this information is authorized to do so.  Therefore, a certificate's issuer MUST do one of the following:
>    - sign the OCSP responses itself, or
>    - explicitly designate this authority to another entity

There is no similar text in the draft. It is thus assumed that the signer of the Status List Tokens are the Issuers and not the List Issuers. 

From a security point of view, using two different entities is better because the key used by the Issuer to issue Referenced Tokens does not need to be communicated or exposed to the Status Issuer. If the Status Issuer sends to the Issuer hash values to be signed by the Issuer, the risk would be that the hash value to be signed does not correspond to a Status List Token but to Referenced Token.

RFC 2560 has been replaced by RFC 6960 precisely to solve the issue of using the same key for two different purposes.

If the keys are different, this indeed complexiflies the PKI, but this path has already be opened when drafting OCSP in particular by defining id-kp-OCSPSigning in an extended key usage certificate extension.

If the List Issuer would use its own key to sign, then it makes sense to consider it as an entity which is different from the Issuer, otherwise it doesn't.

There would be other consequences to consider like [4.2.2.2.1](https://datatracker.ietf.org/doc/html/rfc6960#section-4.2.2.2.1).  Revocation Checking of an Authorized Responder.
What about an extended key usage certificate extension that would be called id-kp-TSLSigning ?

### LINKED ISSUE #234: Validation Rules (section 8.3)
In section 8.3 (Validation Rules), the text states:

> 8.3.  Validation Rules
> 
> The text states:
> 
>    If the validation was successful, the Relying Party MUST
>    perform the following validation steps to evaluate the status of the
>    reference token:
> 
>    (...)
> 
>    3.  Validate the Status List Token:
> 
>        1.  Validate the Status List Token by following the rules defined
>            in section 7.2 of [RFC7519] for JWTs and section 7.2 of
>            [RFC8392] for CWTs
> 

Section 7.2 from RFC 7519 does not say which key SHALL be used to verify the signature of the Status List Token. Section 8.3 from this document is also silent. In order to be able to interoperate, additional information needs to be added into the document. 

The simplest (and the less secure approach) is to use the same private key to sign Referenced Tokens and Status List Tokens. This corresponds to the case where the Status issuer is the Issuer.

A more secure approach is to use of different private keys to sign Referenced Tokens and Status List Tokens. This corresponds to the case where the Status issuer is an entity that has been authorized by the Issuer to issue Status List Tokens. This approach should be RECOMMENDED.

In RFC 5280, a CRL issuer can be nominated by a CA by the issuance of a PKC that contains the cRLSign bit set in the key usage extension (Section 4.2.1.3 from RFC 5280). Since the role of a CRL issuer and of a Status issuer are different, the same bit cannot be used.

In the key usage extension of a PKC, it becomes necessary to define an additional bit that can apply to a Status issuer nominated by an Issuer and to register that additional bit at IANA.

At the moment, in RFC 5280, the following Object Identifier is defined:
{joint-iso-itu-t(2) ds(5) certificateExtension(29) keyUsage(15)}
OID dot notation: 2.5.29.15

```
      id-ce-keyUsage OBJECT IDENTIFIER ::=  { id-ce 15 }
      KeyUsage ::= BIT STRING {
           digitalSignature        (0),
           nonRepudiation          (1), -- recent editions of X.509 have
                                -- renamed this bit to contentCommitment
           keyEncipherment         (2),
           dataEncipherment        (3),
           keyAgreement            (4),
           keyCertSign             (5),
           cRLSign                 (6),
           encipherOnly            (7),
           decipherOnly            (8) }

```
The next usable bit is (9).

The following line should be happened to this list:

`           sLTSign                 (9)`

{joint-iso-itu-t(2) ds(5) certificateExtension(29) keyUsage(15) sLTSign(9)}

OID dot notation: 2.5.29.15.9

With the following explanation:

> The sLTSign bit is asserted when the subject public key can be used for verifying signatures on Status List Tokens.

A specific section dedicated to this extension should be added so that it can be referenced in the IANA registration template.

After these explanations, we can now come back to a text change proposal for section 8.3 (Validation Rules).

Text change proposal for section 8.3.

> 8.3.  Validation Rules
> 
> The text states:
> 
>    If the validation was successful, the Relying Party MUST
>    perform the following validation steps to evaluate the status of the
>    reference token:
> 
>    (...)
> 
>    3.  Identify the public key to be used to verify the signature 
>        of Status List Token
> 
> If no iss claim, nor an x5c claim, is present in the Status List Token and unless a specific local rule applies, the Relying Party MUST verify the Status List Token using the same key that has been used to verify the Referenced Token. This approach is NOT RECOMMENDED.
> 
> If the iss claim is present in the JOSE Header, the JWT Claims, or the COSE Header, a Relying Party can use this parameter to obtain a JSON Web Key to use in the verification process. It should be noticed that, according to RFC 7519, the processing of this claim is generally application specific. Unless the processing of this claim is made in a specific environment or context, this approach is not interoperable.
> 
> If the "x5c" claim is present in the JWT Claims, a Relying Party SHOULD use this parameter to obtain a JSON Web Key to use in the verification process. The "x5c" claim is an array of strings that contains an x.509 certificate chain, in order from the root certificate authority to the signing certificate.
> 
> The last certificate of the chain can be either the certificate of the Issuer or of the Status Issuer. In the later case, the key usage extension (defined in section 4.2.1.3 from RFC 5280) MUST be present and MUST contain an additional bit defined in section 9 of this document. 
> 
> Once it has been verified that the whole chain is valid and that no certificate of the chain has been revoked, the public key from the last certificate of the chain MUST be used to validate the Status List Token. This approach is RECOMMENDED.
> 
> 
> 2.  Validate the Status List Token by following the rules defined in section 7.2 of [RFC7519] for JWTs and section 7.2 of [RFC8392] for CWTs
> 
> 3.  Check for the existence of the required claims as defined in Section 5.1 and Section 5.2 depending on token type

A text proposal for defining the sLTSign bit is provided in a subsequent issue.

### LINKED ISSUE #235: New section to define the sLTSign bit of a key usage extension in a PKC
It is proposed to add a new section 9.

**Status List Token signature bit of a key usage extension in a PKC (sLTSign)**

The key usage extension of a PKC (Public Key Certificate) defines the purpose of the public key contained in the PKC  In RFC 5280, section 4.2.1.3, the text states that "conforming CAs MUST include this extension in certificates that contain public keys that are used to validate digital signatures or other public key certificates or CRLs.  When present, conforming CAs SHOULD mark this extension as critical". At the time RFC 5280 was written and published (2008), the concept of a Status Issuer did not existed.

This document states that conforming CAs MUST include this extension in certificates that contain public keys that are used to validate digital signatures or other public key certificates, CRLs or Status List Tokens.  When present, conforming CAs SHOULD mark this extension as critical.

In RFC 5280, the following Object Identifier has been defined:
{joint-iso-itu-t(2) ds(5) certificateExtension(29) keyUsage(15)}. OID dot notation: 2.5.29.15

 ```
     id-ce-keyUsage OBJECT IDENTIFIER ::=  { id-ce 15 }
      KeyUsage ::= BIT STRING {
           digitalSignature        (0),
           nonRepudiation          (1), -- recent editions of X.509 have
                                -- renamed this bit to contentCommitment
           keyEncipherment         (2),
           dataEncipherment        (3),
           keyAgreement            (4),
           keyCertSign             (5),
           cRLSign                 (6),
           encipherOnly            (7),
           decipherOnly            (8) }

```
In order to allow an Issuer to designate a Status Issuer, an additional bit in the key usage extension is needed. The next usable bit is (9).

The following line is happened to this list:

           sLTSign                 (9)

with the following meaning:

The sLTSign bit is asserted when the subject public key can be used for verifying signatures on Status List Tokens.

The id-ce-keyUsage OBJECT IDENTIFIER ::=  { id-ce 15 } then becomes:

 ```
     id-ce-keyUsage OBJECT IDENTIFIER ::=  { id-ce 15 }
      KeyUsage ::= BIT STRING {
           digitalSignature        (0),
           nonRepudiation          (1), -- recent editions of X.509 have
                                -- renamed this bit to contentCommitment
           keyEncipherment         (2),
           dataEncipherment        (3),
           keyAgreement            (4),
           keyCertSign             (5),
           cRLSign                 (6),
           encipherOnly            (7),
           decipherOnly            (8),
           sLTSign                 (9) }

```

Using the text previously proposed, a section 14.8 (Key Usage extension for Status issuer certificates) should be added into section 14 (IANA Considerations).

### LINKED ISSUE #237: Token Lifecycle
The content of section 13.1 (Token Lifecycle) does not contain sufficient guidance.

The following guidance is rather vague:

>    The lifetime of a Status List Token depends on the lifetime of its eferenced Tokens.

Since the exp (expiration time) claim in a Status List Token is OPTIONAL, it is hard to understand how this sentence can be used.

The three paragraphs do not provide sufficient information

Proposed text replacement:

> The Issuer and the Status issuer SHALL initially agree on the following set of parameters:
> 
> -	the number of bits that indicates the status of each Referenced Token,
> -	the number of entries in a Status List,
> -	the number of Status Lists initially available,
> -  one or more absolute uris where the Status List Tokens will be made available by the Status issuer,
> -	a relative uri associated with each Status List (that can also be used to reference a given Status List).
> 
> The Issuer CAN communicate to the Status issuer the "time to live" which is the maximum amount of time, in seconds, that the Status List Token can be cached by a consumer before a fresh copy SHOULD be retrieved ('ttl' claim). If the "time to live" is communicated, then the Status issuer SHALL include it into each Status List Token.
> 
> For every Status List, the Issuer SHALL communicate to the Status issuer the date when each Status List will cease to be updated by the Issuer. 
> For every Status List, the Issuer CAN communicate to the Status issuer the time at which the Status List Token will be considered expired. If the expiration time is communicated, then the Status issuer SHALL include the 'exp' claim into the Status List Token.
> 
> At any time, the Status issuer and the Issuer may agree to use a new set of parameters. For every Status List already agreed, the agreed parameters remain applicable until the date when the Status List will cease to be updated by the Issuer is reached.
> 
> The Status issuer SHOULD issue and distribute Status List Tokens before the first Referenced Token is issued by the Issuer.
> 
> The status of Referenced Tokens with close expiration times should be grouped together into Status Lists so the corresponding Status List Tokens can be no longer issued and distributed once all the Referenced Tokens referenced in it have expired.
> 

The text continues with :

>    Referenced Tokens may be regularly re-issued to mitigate linkability of presentations to Relying Parties.  In this case, every re-issued
>    Referenced Token MUST have a fresh Status List entry in order to  prevent this becoming possible source of correlation.

The concept of Referenced Tokens that would be regularly re-issued means a same content with a different validity period.

The following sentence looks odd:

>    Referenced Tokens may be regularly re-issued to mitigate linkability of presentations to Relying Parties.  

Since the content of the referenced Token is nearly the same, linkability can easily be achieved and a re-issuance of Referenced Tokens will be unable "to mitigate linkability of presentations to Relying Parties". 

These two sentences should be removed.

The text ends-up with :

>    Referenced Tokens may also be issued in batches, such that Holders can use individual tokens for every transaction.  In this case, every
>    Referenced Token MUST have a dedicated Status List entry.  Revoking batch issued Referenced Tokens might reveal this correlation later
>    on.
> 

The right wording is not "batch issuance of Referenced Tokens" but "one-time-use Referenced Tokens"
The second sentence has nothing to do with a consequence of the previous sentence as it applies to every Referenced Token, whether or not it is a one-time-use Referenced Token.

These three sentences should be removed.

Proposed text replacement:

> One-time-use Referenced Tokens may also be issued in batches, such that Holders can use a different Referenced Token for every transaction.  In this case, the status of every one-time-use Referenced Token SHOULD be handled in a different Status list so that a global revocation of these one-time-use Referenced Tokens cannot be used to reveal a correlation later on.

paulbastian:
Closes #237 
Closes #235
Closes #234 
Closes #227 
Closes #224

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -1399,6 +1399,7 @@ for their valuable contributions, discussions and feedback to this specification
 
 * add considerations about External Status Issuer or Status Provider
 * add recommendations for Key Resolution and Trust Management
+* add extended key usage extensions for x509
 * editorial changes on terminology and Referenced Tokens
 * clarify privacy consideration around one time use reference tokens
 * explain the Status List Token size dependencies
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -870,6 +870,38 @@ A Status List Token in the JWT format should follow the security considerations 
 
 A Status List Token in the CWT format should follow the security considerations of {{RFC8392}}.
 
+## Key Resolution and Trust Management
+
+This specification does not mandate specific methods for key resolution and trust management, however the following recommendations are made:
+
+If the Issuer of the Referenced Token is the same entity as the Status Issuer, then the same key that is embedded into the Referenced Token may be used for the Status List Token. In this case the Status List Token may use:
+- the same `x5c` value or an `x5t`, `x5t#S256` or `kid` parameter referencing to the same key as used in the Referenced Token for JOSE.
+- the same `x5chain` value or an `x5t` or `kid` parameter referencing to the same key as used in the Referenced Token for COSE.
+
+Alternatively, the Status Issuer may use the same web-based key resolution that is used for the Referenced Token. In this case the Status List Token may use:
+- an `x5u`, `jwks`, `jwks_uri` or `kid` parameter referencing to the same key as used in the Referenced Token for JOSE.
+- an `x5u` or `kid` parameter referencing to the same key as used in the Referenced Token for COSE.
+
+If the Issuer of the Referenced Token is a different entity than the Status Issuer, then the keys used for the Status List Token may be cryptographically linked, e.g. by an Certificate Authority through an x.509 PKI. The certificate of the Issuer for the Referenced Token and the Status Issuer should be issued by the same Certificate Authority and the Status Issuer's certificate should utilize [extended key usage](#eku).
+
+~~~ ascii-art
+┌───────────────────────┐   
+│ Certificate Authority │   
+└─┬─────────────────────┘   
+  │ authorize               
+  │  ┌────────┐             
+  ├─►│ Issuer │             
+  │  └─┬──────┘             
+  │    ▼ update status      
+  │  ┌───────────────┐      
+  └─►│ Status Issuer │      
+     └─┬─────────────┘      
+       ▼ provide Status List
+     ┌─────────────────┐    
+     │ Status Provider │    
+     └─────────────────┘    
+~~~
+
 ## Status List Caching
 
 When fetching a Status List Token, Relying Parties must carefully evaluate how long a Status List is cached for. Collectively the `iat`, `exp` and `ttl` claims when present in a Status List Token communicate how long a Status List should be cached and should be considered valid for. The following diagram illustrates the relationship between these claims and how they are designed to influence caching.
@@ -1318,6 +1350,7 @@ for their valuable contributions, discussions and feedback to this specification
 
 -07
 
+* add recommendations for Key Resolution and Trust Management
 * editorial changes on terminology and Referenced Tokens
 * clarify privacy consideration around one time use reference tokens
 * explain the Status List Token size dependencies
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -887,7 +887,7 @@ A Status List Token in the JWT format should follow the security considerations 
 
 A Status List Token in the CWT format should follow the security considerations of {{RFC8392}}.
 
-## Key Resolution and Trust Management
+## Key Resolution and Trust Management {#key-management}
 
 This specification does not mandate specific methods for key resolution and trust management, however the following recommendations are made:
 
@@ -1006,11 +1006,11 @@ To avoid privacy risks for colluding Relying Parties, it is RECOMMENDED that Iss
 
 A Status Issuer and a Relying Party Issuer may link two transaction involving the same Referenced Tokens by comparing the status claims of the Referenced Token and therefore determine that they have interacted with the same Holder. It is therefore recommended to use Status Lists for Referenced Token formats that have similar unlinkability properties.
 
-## Third-Party Hosting {#third-party-hosting}
+## External Status Provider for Privacy {#third-party-hosting}
 
-If the roles of the Issuer and the Status Provider are performed by two different entities, this may give additional privacy assurances as the Issuer has no means to identify the Relying Party or its request.
+If the roles of the Status Issuer and the Status Provider are performed by different entities, this may give additional privacy assurances as the Issuer has no means to identify the Relying Party or its request.
 
-Third-Party hosting may also allow for greater scalability, as the Status List Tokens may be served by operators with greater resources, like CDNs.
+Third-Party hosting may also allow for greater scalability, as the Status List Tokens may be served by operators with greater resources, like CDNs, while still ensuring authenticity and integrity of Token Status List, as it is signed by the Status Issuer.
 
 ## Historical Resolution {#privacy-historical}
 
@@ -1044,6 +1044,14 @@ The storage and transmission size of the Status Issuer's Status List Tokens depe
 The Status List Issuer may increase the size of a Status List if it requires indices for additional Referenced Tokens. It is RECOMMENDED that the size of a Status List in bits is divisible in bytes (8 bits) without a remainder, i.e. `size-in-bits` % 8 = 0.
 
 The Status List Issuer may chunk its Referenced Tokens into multiple Status Lists to reduce the transmission size of an individual Status List Token. This may be useful for setups where some entities operate in constrained environments, e.g. for mobile internet or embedded devices. The Status List Issuer may chunk the Status List Tokens depending on the Referenced Token's expiry date to align their lifecycles and allow for easier retiring of Status List Tokens, however the Status Issuer must be aware of possible privacy risks due to correlations.
+
+## External Status Issuer
+
+If the roles of the Issuer of the Referenced Token and the Status Issuer are performed by different entities, they must align on the key and trust management as described in [](#key-management). This scenario may be necessary or useful if a use case requires that revocations of Referenced Tokens are managed by a different entities, e.g. for regulatory or privacy reasons.
+
+## External Status Provider for Scalability
+
+If the roles of the Status Issuer and the Status Provider are performed by different entities, this may allow for greater scalability, as the Status List Tokens may be served by operators with greater resources, like CDNs. At the same time the authenticity and integrity of Token Status List is still guaranteed, as it is signed by the Status Issuer.
 
 ## Status List Formats
 
@@ -1371,6 +1379,7 @@ for their valuable contributions, discussions and feedback to this specification
 
 -07
 
+* add considerations about External Status Issuer or Status Provider
 * add recommendations for Key Resolution and Trust Management
 * editorial changes on terminology and Referenced Tokens
 * clarify privacy consideration around one time use reference tokens
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -903,14 +903,14 @@ Alternatively, the Status Issuer may use the same web-based key resolution that 
 ┌────────┐  host keys  ┌──────────────────────┐
 │ Issuer ├────────┬───►│ .well-known metadata │
 └─┬──────┘        │    └──────────────────────┘
-  ▼ update status │                           
-┌───────────────┐ │                           
-│ Status Issuer ├─┘                           
-└─┬─────────────┘                             
-  ▼ provide Status List                       
-┌─────────────────┐                           
-│ Status Provider │                           
-└─────────────────┘                           
+  ▼ update status │
+┌───────────────┐ │
+│ Status Issuer ├─┘
+└─┬─────────────┘
+  ▼ provide Status List
+┌─────────────────┐
+│ Status Provider │
+└─────────────────┘
 ~~~
 If the Issuer of the Referenced Token is a different entity than the Status Issuer, then the keys used for the Status List Token may be cryptographically linked, e.g. by an Certificate Authority through an x.509 PKI. The certificate of the Issuer for the Referenced Token and the Status Issuer should be issued by the same Certificate Authority and the Status Issuer's certificate should utilize [extended key usage](#eku).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -870,9 +887,37 @@ A Status List Token in the JWT format should follow the security considerations
...
+- the same `x5c` value or an `x5t`, `x5t#S256` or `kid` parameter referencing to the same key as used in the Referenced Token for JOSE.
+- the same `x5chain` value or an `x5t` or `kid` parameter referencing to the same key as used in the Referenced Token for COSE.
 
-This specification allows for both cryptographic signatures and message authentication codes (MAC) to secure a Status List Token. For (H)MAC based deployments, the security implications of a shared secret should be understood and should match the general security considerations of the deployment. There usually exists a stronger relationship between RP and Issuer for such deployments and they have the capability to securely distribute a shared secret ouf of band.
+Alternatively, the Status Issuer may use the same web-based key resolution that is used for the Referenced Token. In this case the Status List Token may use:
+- an `x5u`, `jwks`, `jwks_uri` or `kid` parameter referencing to the same key as used in the Referenced Token for JOSE.
+- an `x5u` or `kid` parameter referencing to the same key as used in the Referenced Token for COSE.
```
c2bo, line 900:
```suggestion
- an `x5u`, `jwks`, `jwks_uri` or `kid` parameter referencing to a key using the same web-based resolution as used in the Referenced Token for JOSE.
- an `x5u` or `kid` parameter referencing to a key using the same web-based resolution as used in the Referenced Token for COSE.
```

draft-ietf-oauth-status-list.md
```diff
@@ -870,9 +887,37 @@ A Status List Token in the JWT format should follow the security considerations
...
+
```
paulbastian, line 901:
```suggestion

~~~ ascii-art
┌────────┐ resolve key ┌─────────────────────┐
│ Issuer ├────────┬───►│ well-known metadata │
└─┬──────┘        │    └─────────────────────┘
  ▼ update status │                           
┌───────────────┐ │                           
│ Status Issuer ├─┘                           
└─┬─────────────┘                             
  ▼ provide Status List                       
┌─────────────────┐                           
│ Status Provider │                           
└─────────────────┘                           
~~~
```

c2bo:
```suggestion
	~~~ ascii-art
	┌────────┐  host keys  ┌──────────────────────┐
	│ Issuer ├────────┬───►│ .well-known metadata │
	└─┬──────┘        │    └──────────────────────┘
	  ▼ update status │                           
	┌───────────────┐ │                           
	│ Status Issuer ├─┘                           
	└─┬─────────────┘                             
	  ▼ provide Status List                       
	┌─────────────────┐                           
	│ Status Provider │                           
	└─────────────────┘                           
	~~~
```

draft-ietf-oauth-status-list.md
```diff
@@ -807,6 +808,22 @@ The following is a non-normative example for media type `application/json`:
 }
 ~~~
 
+# X.509 Certificate Extensions
+
+## Extended Key Purpose Extension {#eku}
+
+{{RFC5280}} specifies the EKU X.509 certificate extension for use on end entity certificates. The extension indicates one or more purposes for which the certified public key is valid. The EKU extension can be used in conjunction with the Key Usage (KU) extension, which indicates the set of basic cryptographic operations for which the certified key may be used.
```
c2bo, line 815:
```suggestion
{{RFC5280}} specifies the Extended Key Usage (EKU) X.509 certificate extension for use on end entity certificates. The extension indicates one or more purposes for which the certified public key is valid. The EKU extension can be used in conjunction with the Key Usage (KU) extension, which indicates the set of basic cryptographic operations for which the certified key may be used.
```

draft-ietf-oauth-status-list.md
```diff
@@ -807,6 +808,22 @@ The following is a non-normative example for media type `application/json`:
 }
 ~~~
 
+# X.509 Certificate Extensions
+
+## Extended Key Purpose Extension {#eku}
```
c2bo, line 813:
```suggestion
## Extended Key Usage Extension {#eku}
```

draft-ietf-oauth-status-list.md
```diff
@@ -1322,6 +1397,8 @@ for their valuable contributions, discussions and feedback to this specification
 
 -07
 
+* add considerations about External Status Issuer or Status Provider
+* add recommendations for Key Resolution and Trust Management
```
paulbastian, line 1401:
```suggestion
* add recommendations for Key Resolution and Trust Management
* add extended key usage extensions for x509
```

c2bo (approved):

tplooker (approved):

draft-ietf-oauth-status-list.md
```diff
@@ -854,6 +855,69 @@ The following is a non-normative example for media type `application/json`:
...
+                0xC9               0x44               0xF9
+
+~~~
+
+Resulting in the byte array and compressed/base64url-encoded Status List:
+
+~~~~~~~~~~
+{::include ./examples/status_list_encoding2_json}
+~~~~~~~~~~
+
```
c2bo, line 920:
```suggestion
```

PR #249, opened 2025-01-30
222 suspended privacy considerations

### LINKED ISSUE #222: Interims Feedback: Discussion around Suspended Status Type
In the  OAuth Interims Call on 13. Jan 2025, feedback occured towards the pre-defined Status Types.

Currently we define only 3 Status Types:
- 0x00 VALID
- 0x01 INVALID
- 0x02 SUSPENDED
- the rest is undefined or application-specific

There was no objection towards 0x00 and 0x01 and people agreed that for security reasons its best to have those important ones fixed in the specification. Towards 0x02 (suspended) there were different opinions and suggestions. The summarized options are:

## Option A
Keep 0x02 (suspended) as pre-defined but add additional text in the draft that explains potential privacy issues of it (leaking more information than necessary when used with natural persons) but also explains potential use cases (suspended may signal different actions for the Relying Party and avoid deletion of an account or similar)

## Option B
Remove 0x02 (suspended) and have this as an application-specific, ecosystem-depended value, if needed. 

## Option C
Leave 0x00 and 0x01 as pre-defined in the draft and add context-specific values and descriptions for the other ones in the Status List Token, see initial thoughts on this in #1

c2bo:
Closes #222 
Rendered Version: https://drafts.oauth.net/draft-ietf-oauth-status-list/222-suspended-privacy-considerations/draft-ietf-oauth-status-list.html


draft-ietf-oauth-status-list.md
```diff
@@ -973,6 +975,14 @@ By default, this specification only supports providing Status List information f
 
 There are strong privacy concerns that have to be carefully taken into consideration when providing a mechanism that allows historic requests for status information - see [](#privacy-relying-party) for more details. Support for this functionality is optional and Implementers are RECOMMENDED to not support historic requests unless there are strong reasons to do so and after carefully considering the privacy implications.
 
+## Status Types {#privacy-status-types}
+
+As previously explained, there is the danger of observability of Relying Parties and Outsiders. That means that any Status Type that transports special information about a Token can leak information to other parties. This documents defines one additional Status Type with "SUSPENDED" that conveys such additional information. Depending on the use-case, suspended could for example provide information that an authorization in the Token is suspended, but the token itself is still valid.
```
paulbastian, line 980:
```suggestion
As previously explained, there is the potential risk of observability by Relying Parties and Outsiders. That means that any Status Type that transports special information about a Token can leak information to other parties. This documents defines one additional Status Type with "SUSPENDED" that conveys such additional information. Depending on the use-case, suspended could for example provide information that an authorization in the Token is suspended, but the token itself is still valid.
```

draft-ietf-oauth-status-list.md
```diff
@@ -973,6 +975,14 @@ By default, this specification only supports providing Status List information f
 
 There are strong privacy concerns that have to be carefully taken into consideration when providing a mechanism that allows historic requests for status information - see [](#privacy-relying-party) for more details. Support for this functionality is optional and Implementers are RECOMMENDED to not support historic requests unless there are strong reasons to do so and after carefully considering the privacy implications.
 
+## Status Types {#privacy-status-types}
+
+As previously explained, there is the potential risk of observability by Relying Parties and Outsiders. That means that any Status Type that transports special information about a Token can leak information to other parties. This documents defines one additional Status Type with "SUSPENDED" that conveys such additional information. Depending on the use-case, suspended could for example provide information that an authorization in the Token is suspended, but the token itself is still valid.
```
tplooker, line 980: Maybe we should be specific around where this is "previous explained" e.g by linking to a section.

c2bo, line 980:
```suggestion
As previously explained, there is the potential risk of observability by Relying Parties (see [](#privacy-relying-party)) and Outsiders (see [](#privacy-outsider)). That means that any Status Type that transports special information about a Token can leak information to other parties. This documents defines one additional Status Type with "SUSPENDED" that conveys such additional information. Depending on the use-case, suspended could for example provide information that an authorization in the Token is suspended, but the token itself is still valid.
```

tplooker (approved):

PR #270, opened 2025-02-28
Add cddl for statuslist cbor encoding

### LINKED ISSUE #267: CDDL for CBOR
Rohan Mahy suggested on the mailing list:

8. There should be a CDDL snippet for the CBOR. I would be happy to provide one once points 6 and 7 are addressed.

c2bo:
Closes #267 
I only added the CDDL for the StatusList object CBOR encoding which seemed to be the most important to provide CDDL for. I don't think CDDL for the CWT would help (and I didn't find many examples in other specs doing CDDL for CWT).

@rohanmahy Would that work for you?

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -362,7 +362,7 @@ The following is the CDDL definition of the StatusList structure:
 
 ~~~ cddl
 StatusList = {
-    bits: int, ; The amount of bits used per Referenced Token
+    bits: 1 / 2 / 4 / 8, ; The number of bits used per Referenced Token
     lst: bstr, ; Byte string that contains the Status List
     ? aggregation_uri: tstr, ; link to the Status List Aggregation
 }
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -358,6 +358,16 @@ This section defines the data structure for a CBOR-encoded Status List:
...
   * `aggregation_uri`: OPTIONAL. Text string (Major Type 3) that contains a URI to retrieve the Status List Aggregation for this type of Referenced Token. See section [](#aggregation) for further detail.
 
+The following is the CDDL definition of the StatusList structure:
+
+~~~ cddl
+StatusList = {
+    bits: int, ; The amount of bits used per Referenced Token
+    lst: bstr, ; Byte string that contains the Status List
+    ? aggregation_uri: tstr, ; link to the Status List Aggregation
+}
```
rohanmahy, line 368:
```suggestion
StatusList = {
    bits: 1 / 2 / 4 / 8, ; The number of bits used per Referenced Token
    lst: bstr, ; Byte string that contains the Status List
    ? aggregation_uri: tstr, ; link to the Status List Aggregation
}
```

rohanmahy:
> Closes #267 I only added the CDDL for the StatusList object CBOR encoding which seemed to be the most important to provide CDDL for. I don't think CDDL for the CWT would help (and I didn't find many examples in other specs doing CDDL for CWT).
> 
> @rohanmahy Would that work for you?

I proposed a one line change in the CDDL. `bits` can't take any arbitrary integer value. It has exactly 4 choices, so just enumerate them.

Using text string keys instead of integer keys is unconventional. I don't have an objection, as long as you are deliberately making that decision.

I agree that anyone using CWTs does not need you to provide CDDL for a generic structure.

Thanks, -r

paulbastian (approved):

rohanmahy (approved):

PR #296, opened 2025-07-04
change accept header from must to should

### LINKED ISSUE #292: Feedback regarding expected format
Feedback from the mailing list (https://mailarchive.ietf.org/arch/msg/oauth/A6g7CUPZjG7KjozoNrTaEdietWI/): 

> draft-ietf-oauth-status-list offers 2 formats or status list tokens: JWT (JSON Web Token) and CWT (CBOR Web Token). But only provides 1 “uri” field. That’s annoying; not developer-friendly; and unnecessary.
> I suggest defining 2 fields: “jwt_uri” and “cwt_uri”. At least one must be present.
> 1 URI can “work” theoretically, but only if all clients and all servers always use the Accept HTTP request header to do content-negotiation. That complicates all parties. It means you can’t just paste the URI into a browser. You can’t use the simplest HTTP GET method that every programming language offers. Caching … who knows.
Perhaps the worst part is that 1 URI will mostly work even for clients that use a simple get(uri) method and don’t bother about the Accept header. The URI in a JWT will return a JWT (the URI in a CWT will return a CWT). The client will assume the result is what they expect. Then some issuers will require content-negotiation; some clients will break; those clients will be “at fault”, but issuer may need to hack their content-negotiation for interoperability. Better to offer 2 explicit fields for 2 explicit formats."

c2bo:
Closes #292

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -744,7 +744,7 @@ To obtain the Status List Token, the Relying Party MUST send an HTTP GET request
 
 The HTTP endpoint SHOULD support the use of Cross-Origin Resource Sharing (CORS) {{CORS}} and/or other methods as appropriate to enable Browser-based clients to access it, unless ecosystems using this specification choose not to support Browser-based clients.
 
-The Relying Party MUST send the following Accept-Header to indicate the requested response type:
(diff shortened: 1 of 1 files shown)


paulbastian (approved):

tplooker:
Change requested via feedback provided on the mailing list which is referenced in the issue #292. Multiple approvals from both other editors, merging.

PR #299, opened 2025-07-17
fix: issue 298

### LINKED ISSUE #298: property uri
Property `uri` is currently defined as

`The uri (URI) claim MUST specify a String value that identifies the Status List Token containing the status information for the Referenced Token. The value of uri MUST be a URI conforming to [RFC3986].`

The issue here is that an issuer of ATs can and query and path parameters (unique to each issued AT), that would allow to track the usage of ATs. While it's technically impossible to completely "hide" information about who downloaded the status list (I refer to SRC IP, user's agent, etc.), the spec in my view should discourage, and even explicitly forbid such techniques. Otherwise, it leaves the door open for this sort of misuse... which also defects the purpose of being a privacy-preserving way to check the statuses of tokens.

I also think it's a bit better to refer to `uri` as one of properties of the `status` claim, not as `claim`.

tplooker:
Fixes #298

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -1000,7 +1000,7 @@ An Issuer could maliciously or accidentally bypass the privacy benefits of the h
 - Generating a unique Status List for every Referenced Token. By these means, the Issuer could maintain a mapping between Referenced Tokens and Status Lists and thus track the usage of Referenced Tokens by utilizing this mapping for the incoming requests.
 - Encoding a unique URI in each Reference Token which points to the underlying Status List. This may involve using URI components such as query parameters, unique path segments, or fragments to make the URI unique.
 
-This malicious behaviour can be detected by Relying Parties that request large amounts of Referenced Tokens by comparing the number of different Status Lists and their sizes with the volume of Reference Tokens being verified.
+This malicious behavior can be detected by Relying Parties that request large amounts of Referenced Tokens by comparing the number of different Status Lists and their sizes with the volume of Reference Tokens being verified.
 
 ## Observability of Relying Parties {#privacy-relying-party}
 
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -993,9 +993,13 @@ This behaviour may be mitigated by:
 - private relay protocols or other mechanisms hiding the original sender like {{RFC9458}}.
 - using trusted Third Party Hosting, see [](#third-party-hosting).
 
-## Malicious Issuers
+## Issuer Tracking of Reference Tokens
 
-A malicious Issuer could bypass the privacy benefits of the herd privacy by generating a unique Status List for every Referenced Token. By these means, the Issuer could maintain a mapping between Referenced Tokens and Status Lists and thus track the usage of Referenced Tokens by utilizing this mapping for the incoming requests. This malicious behaviour could be detected by Relying Parties that request large amounts of Referenced Tokens by comparing the number of different Status Lists and their sizes.
+A malicious Issuer could bypass the privacy benefits of the herd privacy by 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -993,9 +993,13 @@ This behaviour may be mitigated by:
 - private relay protocols or other mechanisms hiding the original sender like {{RFC9458}}.
 - using trusted Third Party Hosting, see [](#third-party-hosting).
 
-## Malicious Issuers
+## Issuer Tracking of Reference Tokens
 
-A malicious Issuer could bypass the privacy benefits of the herd privacy by generating a unique Status List for every Referenced Token. By these means, the Issuer could maintain a mapping between Referenced Tokens and Status Lists and thus track the usage of Referenced Tokens by utilizing this mapping for the incoming requests. This malicious behaviour could be detected by Relying Parties that request large amounts of Referenced Tokens by comparing the number of different Status Lists and their sizes.
+A malicious Issuer could bypass the privacy benefits of the herd privacy by
```
adeinega, line 998: I’m not entirely sure if it’s always a malicious issuer. This kind of thing can "accidentally" be introduced by a legitimate issuer due to the various reasons (due to lack of recommendations, knowledge, and whatnot).

It's a bit unfortunate that there isn't much RPs can do about it as URLs such as https://example.com/some_tenant_guid1/statuslists/1 and https://example.com/some_tenant_guid2/statuslists/1 are valid ones.

draft-ietf-oauth-status-list.md
```diff
@@ -993,9 +993,13 @@ This behaviour may be mitigated by:
...
 - using trusted Third Party Hosting, see [](#third-party-hosting).
 
-## Malicious Issuers
+## Issuer Tracking of Reference Tokens
 
-A malicious Issuer could bypass the privacy benefits of the herd privacy by generating a unique Status List for every Referenced Token. By these means, the Issuer could maintain a mapping between Referenced Tokens and Status Lists and thus track the usage of Referenced Tokens by utilizing this mapping for the incoming requests. This malicious behaviour could be detected by Relying Parties that request large amounts of Referenced Tokens by comparing the number of different Status Lists and their sizes.
+A malicious Issuer could bypass the privacy benefits of the herd privacy by 
+- Generating a unique Status List for every Referenced Token. By these means, the Issuer could maintain a mapping between Referenced Tokens and Status Lists and thus track the usage of Referenced Tokens by utilizing this mapping for the incoming requests. 
+- Encoding a unique uri in each Reference Token which points to the underlying Status List. This may involve using uri components such as query parameters, unique path segments or fragments to make the uri unique.
```
adeinega, line 1000: Typically, OP specs don't allow the fragment component (#) to be a part of URLs to their enpoints.

Browsers do utilize it, but not the server-side code.

adeinega:
I've also left a comment about the use of claim vs property in my https://github.com/oauth-wg/draft-ietf-oauth-status-list/issues/298.

draft-ietf-oauth-status-list.md
```diff
@@ -993,9 +993,13 @@ This behaviour may be mitigated by:
...
+
+This malicious behaviour can be detected by Relying Parties that request large amounts of Referenced Tokens by comparing the number of different Status Lists and their sizes with the volume of Reference Tokens being verified.
```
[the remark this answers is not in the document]
c2bo: We've been mainly using the British spelling, so I'd keep behaviour as is.

draft-ietf-oauth-status-list.md
adeinega:
```suggestion
- Encoding a unique URI in each Reference Token which points to the underlying Status List. This may involve using URI components such as query parameters, unique path segments, or fragments to make the URI unique.
```

Typically, it's written in uppercase.

PR #303, opened 2025-09-15
add note on CWT encoding

### LINKED ISSUE #302: Clarification : Status List Token (in CWT) to the HTTP Response
I was not able to locate a reference to the specification, with regards to how the a Status List Token in CWT is being included to a HTTP Response.

Is it in hex?

c2bo:
Closes #302 

Wording feels somewhat rough, but I think this is the best place to add that note. Any ideas to improve wording?

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -787,6 +787,8 @@ The HTTP response SHOULD use gzip Content-Encoding as defined in {{RFC9110}}.
 
 If caching-related HTTP headers are present in the HTTP response, Relying Parties MUST prioritize the exp and ttl claims within the Status List Token over the HTTP headers for determining caching behavior.
 
+Note that while the examples for Status List Tokens in CWT format in this document are provided in hex encoding, this does not mean that hex encoding is expected for HTTP messages. The expected encoding in an HTTP response containing a Status List Token in CWT format is binary encoding as defined in section 9.2.1 of {{RFC8392}}. This means that the body of an HTTP response contains the raw CBOR bytes.
+
 ## Validation Rules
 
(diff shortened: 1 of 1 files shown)


babisRoutis (approved):
Crystal clear :crystal_ball:

draft-ietf-oauth-status-list.md
```diff
@@ -787,6 +787,8 @@ The HTTP response SHOULD use gzip Content-Encoding as defined in {{RFC9110}}.
 
 If caching-related HTTP headers are present in the HTTP response, Relying Parties MUST prioritize the exp and ttl claims within the Status List Token over the HTTP headers for determining caching behavior.
 
+Note that while the examples for Status List Tokens in CWT format in this document are provided in hex encoding, this does not mean that hex encoding is expected for HTTP messages. The expected encoding in an HTTP response containing a Status List Token in CWT format is binary encoding as defined in section 9.2.1 of {{RFC8392}}. This means that the body of an HTTP response contains the raw CBOR bytes.
```
paulbastian, line 790: Does it make more sense to add this to the paragraph of line 784?

c2bo: I was thinking about that initially as well since it directly fits to the media type, but it kinda made sense to me to not put it in between the normative statements and instead as a note  afterwards.

paulbastian:
We should also clarify, that the CWT/JWT goes into the HTTP response body?

c2bo:
> We should also clarify, that the CWT/JWT goes into the HTTP response body?

That is why i added this sentence - I guess we could add the JWT one?

> This means that the body of an HTTP response contains the raw CBOR bytes.

c2bo:
Moved the whole note up - seemed more logical if we already state something about the body of the http response here.

paulbastian (approved):

tplooker (approved):

PR #306, opened 2025-10-20
Update tooling again

c2bo:
- Updates gh workflows to most recent version
- Fixes bug with relative ::include paths breaking the release pipeline

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -351,13 +351,13 @@ This section defines the data structure for a JSON-encoded Status List:
 The following example illustrates the JSON representation of the Status List with `bits` 1 from the example above:
 
 ~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


c2bo:
> How in the world did you figure this out! Also why are my older ones working fine still?

it only breaks if you are including with `./<some_path>` - it breaks with a leading `.` and I would assume you don't have that in your includes?

This shellscript here seems to be the one that breaks: https://github.com/martinthomson/i-d-template/blob/00a2d981cb9fde43aae6caca4d69737bdf4dc23f/build-targets.sh#L105. Changing to normal include paths fixed the copy for me - I could see empty files being created otherwise in the folder that gets created by the script.

aaronpk:
Nice, yeah I think I only have like 1 or 2 drafts that even use an include in the first place. Glad you got it sorted!

PR #310, opened 2025-11-11
Add asn.1 module

### LINKED ISSUE #309: Provide ASN.1 module
Feedback regarding IANA registration of Extended Key Usage OID:

> I think that a key purpose for OAuth signing is a good idea, but the I-D should include an ASN.1 module.

> See Appendix A of https://datatracker.ietf.org/doc/draft-jpfiset-lamps-attestationkey-eku/ for an example.

c2bo:
Closes #309 

Preview: https://drafts.oauth.net/draft-ietf-oauth-status-list/309-asn1-module/draft-ietf-oauth-status-list.html


seanturner (changes requested):
Just minor nits!

PR #313, opened 2025-11-19
308 wording mdoc cwt

### LINKED ISSUE #308: Wording section "Referenced token in COSE"
While working on an implementation of the latest draft, I was confused about how to include a referenced token in ISO mdoc. The section starts of with:

> The Referenced Token MAY be encoded as a "COSE Web Token (CWT)" object ...

First of all, this must be a typo _CBOR_ Web Token (see https://www.rfc-editor.org/rfc/rfc8392). But then, after showing an example of a CWT, it seamlessly transitions into a description of how to include the status parameter in the MSO:

> ISO mdoc may utilize the Status List mechanism by introducing the status parameter in the Mobile Security Object (MSO) as specified in Section 9.1.2. The status parameter uses the same encoding as a CWT as defined in Section 6.3.

Here, there is no Section 9.1.2 in this draft, but it also says the CWT encoding is used. However, what follows is an example of an IssuerAuth with a `status` field _not_ using the CWT encoding just described.

For now, I went with what the last example shows, as ISO mdoc doesn't use CWT encoding (to my knowledge). I feel like these should either be two (sub)sections, or the first part (describing the CWT encoding) is not needed at all.

Please let me know if anything I wrote is unclear.

c2bo:
Closes #308

Rendered preview: https://drafts.oauth.net/draft-ietf-oauth-status-list/308-wording-mdoc-cwt/draft-ietf-oauth-status-list.html


draft-ietf-oauth-status-list.md
```diff
@@ -556,15 +556,19 @@ The resulting payload of the example above:
...
+The Referenced Token MAY be encoded as a "CBOR Web Token (CWT)" object according to {{RFC8392}} or other formats based on COSE. Referenced Tokens in CBOR should share the same core data structure for a status list reference:
 
-The following content applies to the CWT Claims Set:
-
-* `65535` (status): REQUIRED. The status claim is encoded as a `Status` CBOR structure and MUST include at least one data item that refers to a status mechanism. Each data item in the `Status` CBOR structure comprises a key-value pair, where the key must be a CBOR text string (Major Type 3) specifying the identifier of the status mechanism and the corresponding value defines its contents. This specification defines the following data items:
+ * The `Status` CBOR structure is a Map that MUST include at least one data item that refers to a status mechanism. Each data item in the `Status` CBOR structure comprises a key-value pair, where the key must be a CBOR text string (Major Type 3) specifying the identifier of the status mechanism and the corresponding value defines its contents.
```
paulbastian:
```suggestion
* The `Status` CBOR structure is a Map that MUST include at least one data item that refers to a status mechanism. Each data item in the `Status` CBOR structure comprises a key-value pair, where the key must be a CBOR text string (Major Type 3) specifying the identifier of the status mechanism and the corresponding value defines its contents.
```

tplooker (approved):

PR #319, opened 2025-12-09
Remove DL suspension example

### LINKED ISSUE #315: Update example around suspension in section 12.8
Currently we offer a concrete example around using the suspension status for driving license suspension which has adverse privacy affects when it is used to suspend driving privileges, there are other better examples of suspension we should update the example to reference instead such as the issuer temporarily suspending an ID because of some risk signal received.

tplooker:
Fixes #315


paulbastian:
Shall we add an entry to document history?

PR #339, opened 2026-01-15
fixes for mike bishops' review

### LINKED ISSUE #335: Mike Bishop's Review
> ### Section 8.4, paragraph 3
> ```
>      To obtain the Status List Token, the Relying Party MUST send an HTTP
>      GET request to the URI provided in the Referenced Token with the
>      additional query parameter time and its value being a unix timestamp.
>      The response for a valid request SHOULD contain a Status List Token
>      that was valid for that specified time or an error.
> 
>      If the Server does not support the additional query parameter, it
>      SHOULD return a status code of 501 (Not Implemented) or if the
>      requested time is not supported it SHOULD return a status code of 406
>      (Not Acceptable).  A Status List Token might be served via static
> ```
> 406 (Not Acceptable) indicates that a server was unable to satisfy a client's
> preferences as expressed through Proactive Negotiation (RFC 9110, Section 12.1),
> but a query parameter isn't that. In fact, HTTP itself has no notion of "query
> parameters" -- the optional query component is merely a part of identifying the
> target resource, and the `key=value` structure is a convention applications can
> use or not (albeit a very common one).
> 
> I would suggest normatively spelling out how you want the query portion of the
> URI constructed and choosing a different status code for the requested resource
> not existing. 404 or 410 might be the most appropriate for a server that can
> provide historical data, but not for the requested timestamp. 501 (Not
> Implemented) seems reasonable for a server that understands the request for a
> historical status list, but does not support such a request, though 404 might be
> equally appropriate as the requested resource does not exist on the server.
> 
> 
> ----------------------------------------------------------------------
> COMMENT:
> ----------------------------------------------------------------------
> 
> ## Comments
> 
> ### Section 3
> 
> Thanks for a good Terminology section. For a crucial word that occurs
> over 1k times in this document, I'm surprised not to see a definition of
> "Status" either in the introduction or here.
> 
> It does finally show up in Section 7!
> 
> ### Section 4, paragraph 1
> 
> Please update the last sentence to also use section numbers.
> 
> ### Section 4.1, paragraph 13
> 
> Step 4 above says these byte arrays will be compressed, but there's no
> mention of that here at the end. While it may not be terribly illustrative,
> it might be worth a sentence confirming that these are run through DEFLATE/ZLIB
> before being used as input to 4.2 or 4.3.
> 
> ### Section 8.1, paragraph 1
> ```
>      The Status Provider SHOULD provide Status List Token on an endpoint
>      serving an HTTP GET request to the URI provided in the Referenced
> ```
> Consider instead:
>    The Status Provider SHOULD return the Status List Token
>    in response to an HTTP GET request to the URI provided in the Referenced
> 
> ### Section 8.1, paragraph 0
> 
> Consider a reference to RFC 9110 for HTTP semantics, Accept, etc.
> 
> ### Section 8.2, paragraph 8
> ```
>      The HTTP response SHOULD use gzip Content-Encoding as defined in
>      [RFC9110] for Status List Tokens in JWT format.
> ```
> Given the variety of content codings which are in use with HTTP, it seems
> strange to call out gzip in particular in this spec. That's not an
> interoperability issue. Consider simply indicating that a content coding *such
> as* gzip could reduce the size of the payload, and reference 9110 for negotiation
> of such codings.
> 
> ### Section 14.2, paragraph 2
> ```
>      Experts.  However, to allow for the allocation of names prior to
>      publication, the Designated Expert(s) may approve registration once
>      they are satisfied that such a specification will be published.
> ```
> Given that Specification Required only requires that the doc exist, not be
> published as an RFC, is this really needed? Does the allocation need
> to happen even before publication as an -00 draft?
> 
> ## Nits
> 
> All comments below are about very minor potential issues that you may choose to
> address in some way - or ignore - as you see fit. Some were flagged by
> automated tools (via https://github.com/larseggert/ietf-reviewtool) so there
> will likely be some false positives. There is no need to let me know what you
> did with these suggestions.
> 
> ### Typos
> 
> #### Section 4.1, paragraph 2
> ```
> -        every operation and thus keeping implementations simpler and less
> -                       ^^^^
> +        every operation, thus keeping implementations simpler and less
> +                       ^
> ```
> 
> #### Section 6.3, paragraph 4
> ```
> -          o  idx: REQUIRED.  Unsigned integer (major type 0) The idx
> +          o  idx: REQUIRED.  Unsigned integer (major type 0). The idx
> +                                                            +
> ```
> 
> #### Section 11.4, paragraph 1
> ```
> -    HTTP clients that follow 3xx (Redirection) class of status codes
> -                                               ---------
> ```
> 
> #### Section 13.2, paragraph 3
> ```
> -    Beware, that this mechanism solves linkability issues between Relying
> -          -
> -    Parties, but does not prevent traceability by Issuers.
> -           -
> ```
> 
> ### Section 1.4, paragraph 1
> ```
>      Representing statuses with bits in an array is a rather old and well-
>      known concept in computer science and there has been prior work to
>      use this for revocation and status management such as a paper by
>      Smith et al. [smith2020let] that proposed a mechanism called
>      Certificate Revocation Vectors based on xz compressed bit vectors for
>      each expiration day and the W3C bit Status List [W3C.SL] that
>      similarly uses a compressed bit representation.
> ```
> This section-as-sentence could be chunked for easier parsing:
> 
> >   Representing statuses with bits in an array is a rather old and well-
> >   known concept in computer science. There has been prior work to
> >   use this for revocation and status management. For example, a paper by
> >   Smith et al. [smith2020let] proposed a mechanism called
> >   Certificate Revocation Vectors based on xz compressed bit vectors for
> >   each expiration day. The W3C bit Status List [W3C.SL]
> >   similarly uses a compressed bit representation.
> 
> ### Grammar/style
> 
> #### "URI", paragraph 6
> ```
> ementers should be particularly careful for the correct parsing and decoding
>                                 ^^^^^^^^^^^
> ```
> The usual collocation for "careful" is "with", not "for". Did you mean "careful
> with"?
> 
> #### Section 9.2, paragraph 1
> ```
> s, since they can be used for denial of service attacks on clients. A client
>                               ^^^^^^^^^^^^^^^^^
> ```
> It appears that hyphens are missing.
> 
> #### Section 12.1, paragraph 1
> ```
> or a query parameter that allows these kind of historic lookups as described
>                                  ^^^^^^^^^^
> ```
> The plural determiner "these" does not agree with the singular noun "kind".
> 
> #### Section 12.4, paragraph 1
> ```
> n rates close to 0% or 100% create lowest sizes while revocation rates close
>                                    ^^^^^^
> ```
> A determiner may be missing.
> 
> #### Section 12.4, paragraph 2
> ```
> 50% and random distribution create highest sizes) * the lifetime of Referenc
>                                    ^^^^^^^
> ```
> A determiner may be missing.
> 
> #### Section 13.5, paragraph 5
> ```
> ., "status_list"). The name is case sensitive. Names may not match other reg
>                                ^^^^^^^^^^^^^^
> ```
> This word is normally spelled with a hyphen.
> 
> #### Section 13.7, paragraph 13
> ```
> ., "status_list"). The name is case sensitive. Names may not match other reg
>                                ^^^^^^^^^^^^^^
> ```
> This word is normally spelled with a hyphen.
> 
> #### Section 14.1.1, paragraph 1
> ```
> meters. The registry records a human readable label, the bit representation a
>                                ^^^^^^^^^^^^^^
> ```
> This word is normally spelled with a hyphen.
> 
> #### Section 14.2, paragraph 2
> ```
>  The name is a human-readable case insensitive label for the Status Type that
>                               ^^^^^^^^^^^^^^^^
> ```
> This word is normally spelled with a hyphen.
> 
> #### Section 14.4.1, paragraph 2
> ```
> 5.7.3). This OID is defined in section Section 10. IANA is requested to regis
>                                ^^^^^^^^^^^^^^^
> ```
> Possible typo: you repeated a word.
> 
> #### Section 16.2, paragraph 15
> ```
> atus List The following example uses a 8-bit Status List (256 possible value
>                                      ^
> ```
> Use "an" instead of "a" if the following word starts with a vowel sound, e.g.
> "an article", "an hour".
> 
> #### "C.4.", paragraph 4
> ```
>  the new ttl claim * clarify the sub claim of Status List Token * relax stat
>                                  ^^^^^^^^^
> ```
> This word is normally spelled as one.
> 
> #### "C.4.", paragraph 4
> ```
>  Change typo in Status List Token sub claim description * Add access token as
>                                   ^^^^^^^^^
> ```
> This word is normally spelled as one.

c2bo:
Closes #335

### DIFF
--- a/draft-ietf-oauth-status-list.md
+++ b/draft-ietf-oauth-status-list.md
@@ -193,7 +193,7 @@ Another possible use case for the Status List is to express the status of verifi
 
 ## Rationale
 
-Revocation mechanisms are an essential part of most identity ecosystems. In the past, revocation of X.509 TLS certificates has been proven difficult. Traditional certificate revocation lists (CRLs) have limited scalability; Online Certificate Status Protocol (OCSP) has additional privacy risks, since the client is leaking the requested website to a third party. OCSP stapling is addressing some of these problems at the cost of less up-to-date data. Modern approaches use accumulator-based revocation registries and Zero-Knowledge-Proofs to accommodate for this privacy gap, but face scalability issues again. Another alternative is short-lived Referenced Tokens with regular re-issuance, but this puts additional burden on the Issuer's infrastructure.
+Revocation mechanisms are an essential part of most identity ecosystems. In the past, revocation of X.509 TLS certificates has been proven difficult. Traditional certificate revocation lists (CRLs) have limited scalability; Online Certificate Status Protocol (OCSP) has additional privacy risks, since the client is leaking the requested website to a third party. OCSP stapling is addressing some of these problems at the cost of less up-to-date data. Approaches based on cryptographic accumulators and Zero-Knowledge-Proofs try to accommodate for this privacy gap, but are currently (in 2026) facing scalability issues and are not yet standardized. Another alternative is short-lived Referenced Tokens with regular re-issuance, but this puts additional burden on the Issuer's infrastructure.
 
 This specification seeks to find a balance between scalability, security and privacy by representing statuses as individual bits, packing them into an array, and compressing the resulting binary data. Thereby, a Status List may contain statuses of many thousands or millions Referenced Tokens while remaining as small as possible. Placing a large number of Referenced Tokens into the same list also offers Holders and Relying Parties herd privacy from the Status Provider.
 
@@ -201,9 +201,8 @@ This specification seeks to find a balance between scalability, security and pri
 
 The decisions taken in this specification aim to achieve the following design goals:
 
-* the specification shall favor a simple and easy-to-understand concept
 * the specification shall be easy, fast and secure to implement in all major programming languages
-* the specification shall be optimized to support the most common use cases and avoid unnecessary complexity of corner cases
+* the specification shall be optimized to support the most common use cases, such as revocation, and avoid unnecessary complexity of corner cases, such as providing multiple statuses for a single token
 * the Status List shall scale up to millions of tokens to support large-scale government or enterprise use cases
 * the Status List shall enable caching policies and offline support
 * the specification shall support JSON and CBOR based tokens
@@ -597,7 +596,7 @@ The following is the CBOR Annotated Hex output of the example above:
 
 ISO mdoc {{ISO.mdoc}} may utilize the Status List mechanism by introducing the `status` parameter in the Mobile Security Object (MSO) as specified in Section 9.1.2 of {{ISO.mdoc}}. The `status` parameter contains the `Status` CBOR structure as described in [](#referenced-token-cose).
 
-It is RECOMMENDED to use `status` for the label of the field that contains the `Status` CBOR structure.
+It is recommended to use `status` for the label of the field that contains the `Status` CBOR structure.
 
 Application of additional restrictions and policies are at the discretion of the Relying Party.
 
@@ -950,9 +949,9 @@ Implementations SHOULD verify correctness using the test vectors given by this s
 
 ## Security Guidance for JWT and CWT
 
-A Status List Token in the JWT format SHOULD follow the security considerations of {{RFC7519}} and the best current practices of {{RFC8725}}.
+A Status List Token in the JWT format MUST follow the security considerations of {{RFC7519}} and the best current practices of {{RFC8725}}.
 
-A Status List Token in the CWT format SHOULD follow the security considerations of {{RFC8392}}.
+A Status List Token in the CWT format MUST follow the security considerations of {{RFC8392}}.
 
 ## Key Resolution and Trust Management {#key-management}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-status-list.md
```diff
@@ -241,6 +241,9 @@ Holder:
 Relying Party:
 : An entity that relies on the Referenced Token and fetches the corresponding Status List Token to validate the status of that Referenced Token. Also known as a Verifier.
 
+Status:
+: The validity state of a Referenced Token as determined by the Issuer of said Token.
```
c2bo:
```suggestion
: A Status describes the current state, mode, condition or stage of an entity that is represented by the Referenced Token as determined by the Status Issuer.
```

draft-ietf-oauth-status-list.md
```diff
@@ -754,7 +759,9 @@ The fetching, processing and verifying of a Status List Token may be done by eit
 
 ## Status List Request {#status-list-request}
 
-The Status Provider SHOULD provide Status List Token on an endpoint serving an HTTP GET request to the URI provided in the Referenced Token, unless the Relying Party and the Status Provider have alternative methods of distribution.
+The default Status List request and response mechanism uses normal HTTP semantics and Content negotiation etc. as defined in {{RFC9110}}.
```
c2bo:
```suggestion
The default Status List request and response mechanism uses HTTP semantics and Content negotiation as defined in {{RFC9110}}.
```

tplooker (approved):

PR #348, opened 2026-01-28
clarify that Status List validation SHOULD not be performed if Refere…

paulbastian:
…nced Token validation is deemed invalid already


c2bo (approved):

c2bo:
Would this work for you as well @OR13 ?

